PR #1, opened 2026-04-30
fix(install): resolve REPO_DIR via fileURLToPath for paths with spaces

lantisprime:
## Summary

- `install.mjs` fails when the repo is cloned into a directory whose absolute path contains a space (e.g. `~/Documents/New project 3/`).
- Root cause: `new URL(import.meta.url).pathname` returns a percent-encoded path (`/Users/x/Documents/New%20project%203/.../install.mjs`). `fs.readdirSync(REPO_SCRIPTS)` then fails because `New%20project%203` isn't a real directory.
- Fix: use `fileURLToPath` from Node's `url` module — the canonical API for converting `import.meta.url` to a filesystem path. Handles spaces, Windows drive letters, and other URL-encoding edge cases.

## Reproduction

```bash
cd ~
mkdir "spaces in path"
cd "spaces in path"
git clone https://github.com/lantisprime/episodic-memory.git
cd episodic-memory
node install.mjs --tool codex
# Before fix: ENOENT on .../spaces%20in%20path/.../scripts
# After fix:  Installed N scripts to ~/.episodic-memory/scripts
```

## Test plan

- [ ] Clone into a path **without** spaces — `node install.mjs --tool codex` still installs cleanly (no regression).
- [ ] Clone into a path **with** spaces — `node install.mjs --tool codex` installs cleanly (was failing before).
- [ ] Clone into a path with other URL-encodable characters (`#`, `&`) — install still works.

### DIFF
--- a/install.mjs
+++ b/install.mjs
@@ -17,10 +17,11 @@
 import fs from 'fs'
 import path from 'path'
 import os from 'os'
+import { fileURLToPath } from 'url'
 
 const GLOBAL_DIR = path.join(os.homedir(), '.episodic-memory')
 const SCRIPTS_DIR = path.join(GLOBAL_DIR, 'scripts')
-const REPO_DIR = path.dirname(new URL(import.meta.url).pathname)
+const REPO_DIR = path.dirname(fileURLToPath(import.meta.url))
 const REPO_SCRIPTS = path.join(REPO_DIR, 'scripts')
 const REPO_INSTRUCTIONS = path.join(REPO_DIR, 'instructions')
 


lantiscooperdev (commented):
## Bot review (substantive findings)

**Scope:** 2-line change to `install.mjs` — adds `import { fileURLToPath } from 'url'` and replaces `new URL(import.meta.url).pathname` with `fileURLToPath(import.meta.url)`.

**Verified:**
- Root cause is correct. `URL.pathname` returns percent-encoded paths per WHATWG URL spec; `fs.*` APIs require decoded filesystem paths.
- `fileURLToPath` is the official Node recommendation (https://nodejs.org/api/url.html#urlfileurltopathurl-options) and supersedes manual decoding. It is also strictly safer than `decodeURIComponent(url.pathname)`, which mishandles Windows paths (`/C:/...` leading slash) and rare edge cases like `%00`.
- Patch applies cleanly to current `main` (488994e); diff is minimal and local to lines that were already the source of the bug.

**Gaps / suggestions (non-blocking):**
1. **No regression test.** Per the convention "regression tests at fix time", a bats/shell test that runs `node install.mjs --tool <x> --project <tmpdir-with-spaces>` and asserts the install succeeds would prevent re-regression. The repo currently has no test harness for `install.mjs`, so adding one in this PR is reasonable scope; alternatively, file a follow-up issue.
2. **Other call sites?** Quickly grepped the repo — `import.meta.url` appears only in `install.mjs`, so this is the only affected file. The scripts in `scripts/` use absolute paths from `os.homedir()` and don't have the same exposure. ✓
3. **No CHANGELOG entry** — repo has no CHANGELOG so N/A.

**Decision:** Findings are minor; recommend the user approve in the GitHub UI to satisfy the `review-required` gate. Test (#1) can be a follow-up.

lantisprime:
Thanks, Claude, for the clean fix here. The `fileURLToPath(import.meta.url)` change handles the spaced-path install case properly.

Issue #192, opened 2026-05-07
plan-gate.sh: blanket mcp__* allow bypasses write-classification


lantisprime:
## Context

Surfaced during Codex review of the permission-prompt reduction plan (Rule 6 second-opinion review on 2026-05-07; finding 4: "Gate-scope refinement must preserve write/destructive enforcement invariants").

## Problem

[hooks/plan-gate.sh:51](https://github.com/lantisprime/episodic-memory/blob/main/hooks/plan-gate.sh#L51) treats every `mcp__*` tool call as read-only and exits 0 without classification:

```bash
case "$TOOL_NAME" in
  Read|Glob|Grep|Agent|WebFetch|WebSearch|AskUserQuestion|EnterPlanMode|ExitPlanMode|ListMcpResourcesTool|ReadMcpResourceTool|Skill|NotebookRead|ToolSearch|mcp__*)
    exit 0
    ;;
esac
```

But MCP tools span the full action spectrum:

- Read-only: `mcp__claude_ai_Mermaid_Chart__validate_and_render_mermaid_diagram`, `mcp__ccd_session_mgmt__list_sessions`, `mcp__mcp-registry__search_mcp_registry`
- **Write/destructive**: `mcp__Claude_in_Chrome__navigate`, `mcp__Claude_in_Chrome__form_input`, `mcp__Claude_in_Chrome__upload_image`, `mcp__Claude_in_Chrome__file_upload`, `mcp__ccd_session__spawn_task`, `mcp__scheduled-tasks__create_scheduled_task`, `mcp__scheduled-tasks__update_scheduled_task`

Under an armed `.plan-approval-pending` marker, plan-gate currently lets all of these through — including external mutations like file uploads to web pages, scheduled-task creation, or session spawning.

## Impact

Severity: medium (no concrete misuse observed, but the contract documented in the hook header — "blocks write tools while marker exists" — is not honored for MCP tools).

## Proposed fix

Replace the blanket `mcp__*` match with explicit classification:

1. Maintain an allowlist of read-only `mcp__*` tools (validated, listed, searched, read).
2. All other `mcp__*` tool calls block under armed marker.
3. Plus: an escape hatch for "I know this is read-only, let me through" e.g. via project-level settings.

Alternative: tag MCP tools by action class via their description metadata, classify dynamically. More work, more robust.

## Out of scope here

[...]

PR #193, opened 2026-05-07
fix(hooks): exempt /dev/null sink redirects from shared_write upgrade

lantisprime:
## Summary

- `hooks/lib/command-classifier.sh` was upgrading any read-only command with a redirect to `/dev/null` (or sibling device sinks) to `shared_write`, causing plan-gate to block defensive `cmd 2>/dev/null` patterns under armed `.plan-approval-pending`. This was the dominant false-positive trigger for permission prompts during planning sessions.
- Fix exempts `/dev/null`, `/dev/stdout`, `/dev/stderr`, `/dev/tty`, `/dev/zero` from the `has_nonmarker_redirect` upgrade. Read-only commands keep `read_only`; write/push/marker/unsafe commands keep their normal classification.
- 22 new regression tests across 6 axes; 186/186 tests pass. Codex review of the diff returned 4 ACCEPT-WITH-MOD findings — all addressed before commit.

## Discovery

User pain: Claude Code permission prompts firing constantly during planning sessions. Plan-gate's read-only carve-out at [plan-gate.sh:51](hooks/plan-gate.sh#L51) and [:67-71](hooks/plan-gate.sh#L67) was supposed to allow read-only Bash through, but the classifier was returning `shared_write` for anything with a `2>/dev/null`. Step-1 investigation traced this to [command-classifier.sh:489-503](hooks/lib/command-classifier.sh#L489) treating every non-marker redirect target as a real write.

## Invariants preserved

| Class | Test | Verdict |
|---|---|---|
| Real-file writes | T228, T229 | still `shared_write` |
| Compound dev-null + real-write | T230 | still `shared_write` (most-restrictive) |
| Push under dev-null | T231 | still `push_or_pr_create` |
| Marker write + sibling dev-null | T232, T233 | still `marker_write` |
| Unsafe (eval, backticks) + dev-null | T234, T235 | still `unsafe_complex` |
| Look-alike paths (/dev/null2, /devnull) | T240, T241 | still `shared_write` |
| Symlink to /dev/null | n/a (exact-string match) | falls through to `shared_write` (conservative) |

## Codex review findings (all addressed)

- F1 ACCEPT-WITH-MOD — completed device-sink class (added stdout/stderr/tty/zero)
- F2 ACCEPT-WITH-MOD — added marker+sink regression tests
- F3 ACCEPT-WITH-MOD — added unsafe+sink regression tests
- F4 ACCEPT-WITH-MOD — corrected docstring on what stays read-only vs. what keeps its class

Residual concern (out of scope): plan-gate's blanket `mcp__*` allow at [plan-gate.sh:51](hooks/plan-gate.sh#L51) bypasses write classification for MCP tools — filed as [#192](https://github.com/lantisprime/episodic-memory/issues/192).

## Test plan

- [x] `bash tests/test-command-classifier.sh` — 186 passed, 0 failed
- [x] E2E: install updated classifier globally; confirm originally-blocked `ls 2>/dev/null` chain now passes plan-gate
- [x] E2E: confirm `echo > /tmp/foo` still blocked under armed marker
- [x] E2E: confirm `git push` still blocked under armed marker
- [ ] Reviewer: spot-check that no test was relaxed (only additions)

### DIFF
--- a/hooks/lib/command-classifier.sh
+++ b/hooks/lib/command-classifier.sh
@@ -484,12 +484,27 @@ _classify_segment() {
   # If any output redirect targets a non-marker → has_nonmarker_redirect, which
   # forces shared_write later (overrides read-only command classification —
   # `echo hello > file.txt` is a write even though `echo` is read-only).
+  #
+  # Exception: device pseudo-files in /dev/ that are non-persistent sinks/sources
+  # (/dev/null, /dev/stdout, /dev/stderr, /dev/tty, /dev/zero) never affect
+  # shared state, so they must NOT upgrade the redirect-source command. A
+  # read-only command such as `ls >/dev/null` remains read_only; other commands
+  # keep their normal classification (e.g. `git push >/dev/null` still
+  # classifies as push_or_pr_create via its own command rule). Exact-string
+  # match — symlinks to /dev/null fall through to the conservative shared_write
+  # default.
   local r
   local has_nonmarker_redirect=0
   for r in ${REDIRS[@]+"${REDIRS[@]}"}; do
     local rop="${r%%	*}"
     local rtarget="${r#*	}"
     local rbase="$(basename "$rtarget")"
+    case "$rtarget" in
+      /dev/null|/dev/stdout|/dev/stderr|/dev/tty|/dev/zero)
+        # Benign device sink — skip the has_nonmarker_redirect upgrade.
+        continue
+        ;;
+    esac
     case "$rbase" in
       .pre-checkpoint-done|.post-checkpoint-done|.plan-approval-pending)
         local abs_target
--- a/tests/test-command-classifier.sh
+++ b/tests/test-command-classifier.sh
@@ -289,6 +289,51 @@ assert_label "T207 gh pr update-branch" "gh pr update-branch 113" "push_or_pr_cr
 assert_label "T208 gh pr revert" "gh pr revert 113" "push_or_pr_create"
 
 echo ""
+echo "--- /dev/null sink (issue: 2>/dev/null false-positive) ---"
+# Bug: any output redirect to a non-marker forced shared_write, even when
+# the redirect target was /dev/null (the universal sink). This made every
+# defensive `cmd 2>/dev/null` block under plan-gate, which was the dominant
+# false-positive trigger for permission prompts during planning sessions.
+# Fix: /dev/null targets bypass the has_nonmarker_redirect upgrade.
+assert_label "T220 ls 2>/dev/null" "ls /tmp 2>/dev/null" "read_only"
+assert_label "T221 ls >/dev/null" "ls /tmp >/dev/null" "read_only"
+assert_label "T222 ls &>/dev/null" "ls /tmp &>/dev/null" "read_only"
+assert_label "T223 cat 2>/dev/null" "cat /etc/hosts 2>/dev/null" "read_only"
+assert_label "T224 git status 2>/dev/null" "git status 2>/dev/null" "read_only"
+assert_label "T225 grep with stderr sink" "grep -r foo . 2>/dev/null" "read_only"
+# Compound: each segment classified independently, /dev/null exception applies.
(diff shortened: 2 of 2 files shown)


lantiscooperdev (commented):
## Bot review (independent pass over the diff)

Reviewed [hooks/lib/command-classifier.sh](hooks/lib/command-classifier.sh) and [tests/test-command-classifier.sh](tests/test-command-classifier.sh) against the stated invariants and the prior Codex round.

### What I checked

1. **Class completeness of benign sinks.** The five exempted paths (`/dev/null`, `/dev/stdout`, `/dev/stderr`, `/dev/tty`, `/dev/zero`) cover the standard non-persistent device targets. `/dev/random` and `/dev/urandom` are *read* devices — redirecting **out** to them is meaningless but technically allowed; they were correctly not added (writes there fail at runtime, but the classifier shouldn't have to model that).

2. **Tokenizer behavior with the new exemption.** The `continue` lands inside the `for r in REDIRS` loop in `_classify_segment`, after `rop`/`rtarget`/`rbase` are populated. It bypasses both the marker-write `case` and the `*) has_nonmarker_redirect=1` arm. Subsequent iterations re-enter the loop body cleanly. No state leaks across iterations. Sound.

3. **Order-independence.** The marker-write check returns immediately (`return 0`) whenever it matches, so a sibling `2>/dev/null` redirect *before* the marker write doesn't mask it. T232/T233 pin this in both orderings — good.

4. **Exact-string match (no symlink resolution).** This is the right call for a hook script: classifier can't safely call `realpath` on hot paths under a non-trivial perf/portability budget, and a symlinked `/tmp/foo → /dev/null` falling through to `shared_write` is the conservative direction.

5. **Test signal-to-noise.** The 22 new tests are tightly scoped — each row tests exactly one mechanism (sink class, sibling marker, sibling unsafe, look-alike path, compound most-restrictive). No "vibes" tests. T240/T241 (`/dev/null2`, `/devnull`) are particularly nice because they pin the exact-string match against drift.

### Findings

**N1 (NIT, optional):** The block comment at [command-classifier.sh:486-495](hooks/lib/command-classifier.sh#L486) now contains both the original "non-marker forces shared_write" rule and the new exception. Consider extracting the device-sink list to a small `_is_benign_sink()` helper if more sinks need adding later — keeps the redirect loop tight and makes the policy testable independently. Not blocking; the inline `case` is fine for 5 entries.

**N2 (NIT, optional):** Test names T220-T241 are clear, but `T220 ls 2>/dev/null` and `T221 ls >/dev/null` differ only in the operator. A reader scanning failures would benefit from `T220 stderr-to-devnull` / `T221 stdout-to-devnull` style — current names are descriptive enough though.

**No blockers. No correctness concerns. No invariant regressions.**

The fix targets the right code path. The expanded test matrix correctly proves the invariants Codex flagged in the planning-phase review (write-still-blocks, push-still-wins, marker-still-detected, unsafe-still-wins). E2E verification against the installed copy was performed in-PR and documented. [Issue #192](https://github.com/lantisprime/episodic-memory/issues/192) appropriately captures the deferred `mcp__*` concern instead of expanding scope here.

Approve the PR in the GitHub UI when ready (per Rule 17, this comment is the bot review; user approval is the gate).

lantisprime:
Independent Codex review (third pass, after offline diff review and bot review) surfaced one real finding — a pre-existing latent bug the redirect block touches but doesn't introduce or fix.

**Finding:** marker-write redirect early-returns without inspecting sibling redirects, so `echo ok > .claude/.plan-approval-pending > /tmp/foo` masks the `/tmp/foo` write.

**Decision:** filed as #194; not addressing in this PR (out of scope; pre-existing; separately well-defined).

Codex's other focus areas (shell-expansion bypass, cross-hook impact on checkpoint-gate, mixed `/dev/null` + real-file orderings, spec compliance with planning-review finding 4) — all clean. No correctness concerns with the `/dev/null` change itself.

lantiscooperdev (approved):

Issue #226, opened 2026-05-11
transcript-walker.mjs: EACCES on one file aborts entire mining run


lantisprime:
## Summary

`scripts/lib/transcript-walker.mjs:178` opens each `.jsonl` with `fs.createReadStream` inside the main `for await` loop with no per-file try/catch. One EACCES on any unreadable file kills the whole walk, producing 0 candidates.

## Repro

On macOS, currently 4 transcript files in `~/.claude/projects/-Users-charltond-ho-Developer-projects-episodic-memory/` are owned by `root:staff` mode 600 (timestamps cluster 2026-05-10 08:54–09:04, likely from a sudo'd claude invocation):

```
-rw-------  1 root staff   4235  10 May 08:59 733b5790-….jsonl
-rw-------  1 root staff  87084  10 May 08:59 d38f1d5f-….jsonl
-rw-------  1 root staff  14979  10 May 08:54 ed3e1263-….jsonl
-rw-------  1 root staff   7543  10 May 09:04 ed80f4b2-….jsonl
```

Run `node scripts/em-mine-transcripts.mjs --slug episodic-memory --since "$(date -v-1d -u +%Y-%m-%dT00:00:00Z)"` → exits non-zero on first EACCES; output file never created.

## Surfaced by

Wrapper-fix work (launchd routines `--`-separator fix) E2E run on 2026-05-11. Claude inside the routine correctly diagnosed the bug at runtime — but daily-mining still produced 0 candidates because of this walker behavior.

## Class

Same class as the existing `try { fs.statSync... } catch { continue }` at `transcript-walker.mjs:63` — needs per-file fault tolerance one level deeper.

## Proposed fix

Wrap the `createReadStream` + `readline` consumption per-file in try/catch; on EACCES (and arguably any IO error), emit a warning to stderr with the file path + error code and `continue` to the next file. Treat read failure as 0 contribution, not a fatal abort.

## Mitigation today

`sudo chown charltond.ho:staff <four files>` unblocks today's mining without touching the walker.

## Discovered via

User-pref Rule 18 step 8 (E2E testing during launchd routines wrapper-fix). Filed per inline-FU heuristic (architectural / cross-file → GitHub Issue, not inline).

Issue #221, opened 2026-05-10
second-opinion: extend snapshot provider shape validation (cli_match, binary, agent_block_patterns, agent_allow_patterns) — fail-closed in hook (F2, P2)


lantisprime:
## Summary

The provider registry validator (`scripts/second-opinion/lib/registry-validator.mjs`) only validates `id` and `prompt_max_chars`. The hook (`hooks/second-opinion-gate.mjs`) reads additional fields from the installed snapshot — `cli_match`, `binary`, `agent_block_patterns`, `agent_allow_patterns` — without validating them, and `readSnapshot()` (`scripts/second-opinion/lib/install-snapshot.mjs:61-89`) only requires `source_hash` to be present. Empty `providers: []` and malformed `cli_match` regexes both **fail open** for direct provider-shaped tool calls.

Source: codex PR-level review of harness commits 1-6, reply episode `20260510-090140-hold-pr-level-second-opinion-harness-com-59d6`, finding F2 (P2, NEW-CONCERN).

## Repro (codex-executed)

```text
snapshot providers: []
input: Bash codex exec "review", run_in_background:true
observed hook stdout: empty allow

snapshot providers: [{ id:"codex", cli_match:"[", prompt_max_chars:200000, ... }]
input: Bash codex exec "review", run_in_background:true
observed hook stdout: empty allow
```

The second case is a **direct bypass** of block class 1 (`run_in_background`) when the snapshot is syntactically JSON-valid but semantically invalid. Same class as the requested commit-1 registry validator class-completeness audit.

## Root cause

Validator scope is too narrow. `validateProviderRegistry()` was implemented to cover the v3.3 plan's stated invariants (`id` non-empty unique + `prompt_max_chars` non-negative integer), but the hook's block matrix depends on additional provider-shape fields that were never added to the contract. The hook's `compileCliMatch()` returns `null` on invalid regex, and the for-loop `if (!cliRe || !cliRe.test(command)) continue` silently skips that provider — there is no fail-closed branch.

`readSnapshot()` only validates the snapshot envelope (`source_hash` exists), not the provider entries inside `snapshot.providers[]`. There is no shared validator between `register at install time` and `read at hook time`.

## Required fix

1. **Extend `validateProviderRegistry()`** in `scripts/second-opinion/lib/registry-validator.mjs` to enforce per-provider:
   - `cli_match` — non-empty string + valid regex (compile via `new RegExp` in try/catch)
   - `binary` — non-empty string
[...]

PR #228, opened 2026-05-11
fix(launchd): wrapper-mechanism for SKILL-driven routines (Unknown command)

lantisprime:
## Summary

Fixes the launchd-scheduled routines `em-daily-mining`, `em-weekly-digest`, and `instruction-hygiene`, which were 100% broken since `install-launchd-routines.sh` shipped — every invocation produced `Unknown command: /<skill-name>` because the plists called `claude -p /<name>` as a slash command, but those names are SKILLs under `~/.claude/scheduled-tasks/`, not registered slash commands.

This change replaces the broken slash-command invocation with a parameterized wrapper (`em-skill-wrapper.sh`) that inlines `SKILL.md` content as the prompt to `claude -p`, with project-root hard-binding, install-time-pinned absolute binary, plist env defense for the SessionStart handoff guard, and sed-substitution safety.

## Failure mode (before)

`~/Library/Logs/episodic-memory/em-daily-mining.log` (every run since install):
```
Unknown command: /episodic-memory-daily-mining
Unknown command: /episodic-memory-daily-mining
```

`install-launchd-routines.sh:108` emitted `<string>/$1</string>` assuming SKILLs auto-resolve as slash commands. They don't — `~/.claude/scheduled-tasks/` is not a Claude command source.

## What's in this PR

**5 files, +653/−19:**

| File | Change |
|---|---|
| `scripts/em-skill-wrapper.sh.tmpl` | new (44 lines) — wrapper template with `@CLAUDE_BIN@` + `@PROJECT_DIR@` placeholders; `cd PROJECT_DIR` before exec; uses `--` separator before the prompt so SKILL.md's YAML frontmatter (`---`) doesn't trip claude's arg parser |
| `scripts/lib/render-input-validation.sh` | new (37 lines) — sourceable deny-list validator (rejects `\|`, `\\`, `&`, newline in render inputs) |
| `scripts/lib/render-input-validation-test.sh` | new (30 lines) — unit test, 1 positive + 8 negative cases all pass |
| `scripts/lib/launchd-routines-e2e-test.sh` | new (466 lines) — 15-test E2E suite covering wrapper-mechanism, plist contents, cwd-binding, PATH-poison, HOME-elsewhere, worktree caller, nested-cwd, validation-lib unit, dry-run, uninstall fail-open, rollback |
| `install-launchd-routines.sh` | modified — defers preflight to install branch (uninstall fail-open); fixes `gh` preflight from `command -v gh` to `command -v gh \|\| true` so `set -e` doesn't kill on absent `gh`; sources + invokes validator; renders wrapper with sed + sentinel-check; emits `/bin/bash <wrapper> <skill>` instead of `/<skill>`; adds `CLAUDE_SCHEDULED_TASK=1` to plist env (defense for `session-handoff-prompt.sh` env-var guard); `--uninstall` removes rendered wrapper |

## Review process (per Rule 18)

**Plan review:** 7 rounds of codex consensus on `scripts/second-opinion.mjs` harness.
- Trajectory: R1=4 → R2=3 → R3=6 → R4=3 → R5=3 → R6=3 → R7=**ACCEPT**
- Episodes archived in `.episodic-memory/episodes/` (`20260511-07*-launchd-routines-fix-r*` request/reply chain).

**Impl review:** 1 round, codex returned **ACCEPT-with-FU**. All 3 P2 follow-ups applied inline before push:
- positive cwd-binding stub-claude test (`T4a`)
- `command -v gh \|\| true` fix (production bug — `set -e` would kill the script when `gh` absent)
- validator docstring rewritten to reflect actual deny-list semantics (was claiming "strict allowlist")

**E2E:** 15/15 pass against the installed system.

```
T1   dry-run / no persistent writes                  PASS
T2   uninstall fail-open w/ stubbed bins             PASS
T3   install + wrapper content                       PASS
T4a  positive cwd-binding via stub Claude            PASS
T4   artifact-location, /tmp caller (negative)       PASS
T5   detached-worktree caller                        PASS
T6   HOME-elsewhere                                  PASS
T7   PATH-poison (binary pinning)                    PASS
T8   stub-Claude argv contract                       PASS
T9   validation-lib unit (1 pos + 8 neg)             PASS
T10  launchctl print all 3 wrapped                   PASS
T11  kickstart smoke, daily-mining                   PASS
T12  nested-cwd from <project>/scripts               PASS
T13  rollback (uninstall+reinstall)                  PASS
T14  plist content verification (3 plists)           PASS
```

## Dependency note

This branch is currently based on `bccb7fc` (commit 4 of #221 `fix/issue-221-snapshot-validator-extension`), not directly on `main`. The 4 #221 commits ahead of `main` will appear in this PR's "Files changed" tab even though they belong to #221.

**After #221 merges**, this PR can be rebased onto main with `git rebase main fix/launchd-routines-unknown-command` and the duplicate commits will drop out automatically (clean fast-forward).

If review priority is reversed, this PR's content can also be cherry-picked onto a fresh branch off `main` — the routine-fix touches zero files that #221 touches (`install-launchd-routines.sh` vs `install.mjs` etc.).

## Deferred follow-up

#226 — `scripts/lib/transcript-walker.mjs:178` lacks per-file try/catch, so one EACCES (e.g. root-owned `.jsonl` from a sudo'd claude run) aborts the entire walk. Surfaced by this PR's E2E because the routine finally ran far enough to expose it. Out of scope here; the wrapper-mechanism fix is independent and stable regardless.

## Test plan

- [x] `bash install-launchd-routines.sh --dry-run` — renders plists + wrapper without writing
- [x] `bash install-launchd-routines.sh` — installs cleanly; `launchctl print gui/$UID/com.charltonho.em-daily-mining` shows wrapper + `CLAUDE_SCHEDULED_TASK=1`
- [x] `bash scripts/lib/launchd-routines-e2e-test.sh` — 15/15 green
- [x] `launchctl kickstart -kp gui/$UID/com.charltonho.em-daily-mining` — log shows no "Unknown command", claude reaches SKILL execution
- [x] `bash install-launchd-routines.sh --uninstall` — removes all plists + rendered wrapper; repo template untouched
- [x] Uninstall fail-open verified with stubbed PATH (only `/tmp/stubs:/bin:/usr/bin`, no claude/node on PATH)

### DIFF
--- a/hooks/lib/registry-validator.mjs
+++ b/hooks/lib/registry-validator.mjs
@@ -1,0 +1,1 @@
+../../scripts/second-opinion/lib/registry-validator.mjs
--- a/hooks/second-opinion-gate.mjs
+++ b/hooks/second-opinion-gate.mjs
@@ -42,6 +42,8 @@ import fs from 'node:fs'
 import path from 'node:path'
 import os from 'node:os'
 
+import { validateProviderRegistry } from './lib/registry-validator.mjs'
+
 const SNAPSHOT_PATH = process.env.SO_INSTALL_SNAPSHOT_PATH ||
   path.join(os.homedir(), '.claude', 'hooks', 'second-opinion-providers.json')
 
@@ -93,6 +95,19 @@ function loadSnapshot() {
   }
   if (!parsed.source_hash) {
     return { error: 'snapshot-missing-source-hash' }
+  }
+  // I-NEW-B: validate providers[] shape so the hook fail-closes on bad
+  // registry rather than silently skipping providers with bad regex via
+  // compileCliMatch's null-return fallback.
+  try {
+    validateProviderRegistry({ schema_version: 1, providers: parsed.providers })
+  } catch (e) {
+    return {
+      error: 'snapshot-invalid-providers',
+      detail: e.message,
+      field: e.field,
+      provider: e.provider,
+    }
   }
   return { snapshot: parsed }
 }
@@ -201,10 +216,18 @@ if (!input) {
 
 const snap = loadSnapshot()
 if (snap.error) {
+  const extra = { code: snap.error }
+  if (snap.field) extra.field = snap.field
+  if (snap.provider) extra.provider = snap.provider
+  if (snap.detail) extra.detail = snap.detail
+  const detailSuffix = snap.field
+    ? ` (field: ${snap.field}${snap.provider ? `, provider: ${snap.provider}` : ''})`
+    : ''
   emitBlock(
-    `second-opinion-gate: ${snap.error}. Run: node install.mjs --install-second-opinion to install the registry. ` +
-    `(Direct provider calls cannot be safely gated without the install snapshot.)`,
-    { code: snap.error }
+    `second-opinion-gate: ${snap.error}${detailSuffix}. ` +
+    `Run: node install.mjs --install-second-opinion to install/refresh the registry. ` +
+    `(Direct provider calls cannot be safely gated without a valid install snapshot.)`,
+    extra
   )
 }
 
--- a/install-launchd-routines.sh
+++ b/install-launchd-routines.sh
@@ -14,6 +14,7 @@
 # Reversible: --uninstall undoes everything this script created.
 
 set -e
+set -o pipefail
 
 DRY_RUN=0
 SMOKE=0
@@ -32,16 +33,13 @@ LA_DIR="$HOME/Library/LaunchAgents"
 LOG_DIR="$HOME/Library/Logs/episodic-memory"
 PROJECT_DIR="$HOME/Developer/projects/episodic-memory"
 SCHEDULED_TASKS_DIR="$HOME/.claude/scheduled-tasks"
+WRAPPER_TMPL="$PROJECT_DIR/scripts/em-skill-wrapper.sh.tmpl"
+WRAPPER_OUT="$SCHEDULED_TASKS_DIR/em-skill-wrapper.sh"
+VALIDATION_LIB="$PROJECT_DIR/scripts/lib/render-input-validation.sh"
 
-# Pin absolute binaries at install time (round-3 fix: don't rely on PATH lookup at runtime)
-NODE_BIN=$(command -v node)
-GH_BIN=$(command -v gh)
-CLAUDE_BIN=$(command -v claude)
-
-if [ -z "$NODE_BIN" ] || [ -z "$CLAUDE_BIN" ]; then
-  echo "ERROR: node and claude must be on PATH at install time. Found: node=$NODE_BIN claude=$CLAUDE_BIN" >&2
-  exit 3
-fi
+# Binary + template resolution deferred until install branch so --uninstall
+# remains fail-open: it must remove known runtime artifacts without requiring
+# node/claude/template/skill files to be present.
 
 # Plist labels
 LABELS=(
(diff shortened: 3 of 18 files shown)


lantiscooperdev (commented):
## Review (independent voice, per Rule 17)

Bot reviewer here. This PR closes the long-standing "Unknown command" failure on `em-daily-mining`/`em-weekly-digest`/`instruction-hygiene`. Findings below — none blocking, two worth confirming before merge.

### Strengths

1. **Right diagnosis.** The slash-command-vs-SKILL confusion (`install-launchd-routines.sh:108`) is the actual root cause; the env-var defense (Layer B) is a legitimate but secondary layer. Initial scoping correctly separated them.
2. **Test coverage is solid for the scope claimed.** 15 E2E cases hit cwd-binding (4 axes: /tmp, detached worktree, nested subdir, HOME-elsewhere), binary pinning, argv contract, fail-open uninstall, plist content exactness, validation-lib decision fan-out. T4a's positive cwd-binding via stub-claude is the strongest single test in the suite.
3. **Catches a real production bug as a side effect:** `command -v gh || true`. Without that, any install on a machine without `gh` would silently exit at preflight under `set -e`. P1 in disguise.
4. **`--` separator was the right call** for the YAML-frontmatter edge case. Easy to miss in a plan-time review; good that the E2E surfaced it before merge.
5. **Defense-in-depth on the wrapper render path** is well-layered: validator chokepoint + post-render sentinel grep + atomic mv + chmod. Each layer catches a different class.

### Considerations (non-blocking)

1. **Validator scope honesty.** The R5→R6 docstring rewrite in `scripts/lib/render-input-validation.sh` is good — explicitly calls out that shell-sensitive chars (`"`, `$`, backtick, `;`) are NOT rejected, and points to `printf %q` as the alternative if hostile-input defense is needed. This is the right scope for a user-run installer trust boundary. **Recommend keeping the comment as-is** even if a reviewer suggests tightening; the deny-list / allowlist tradeoff is documented honestly.

2. **`d5a8ccf`/`5e13055` cherry-pick noise.** The branch was at one point carrying a duplicate of #221 commit 5 (cherry-picked into both branches), then reset out of this one. Reflog has the trail. **No content lost** — `5e13055` is on `fix/issue-221-snapshot-validator-extension`. Worth flagging here because if reviewers check the reflog they'll see the reset.

3. **Branch base.** PR is based on `bccb7fc` (#221 commit 4), not `main`. The 4 #221 commits appear in this PR's "Files changed" tab as extra noise. Easy to filter — they don't touch any files this PR touches (`install-launchd-routines.sh` vs `install.mjs` etc.). After #221 merges, `git rebase main` here will be a clean fast-forward.

4. **Coverage of weekly-digest/instruction-hygiene execution-correctness.** Out of scope per the codex-approved R5 plan. T14 (plist content verification) covers wrapper-mechanism delivery for all 3 plists; SKILL behavior for the global-store-writing SKILLs is intentionally deferred. **The dependency note in the PR body documents this correctly.**

### Verification I ran

- Walked `scripts/em-skill-wrapper.sh.tmpl` line-by-line; confirmed (a) `cd "@PROJECT_DIR@"` precedes `exec`, (b) `--` precedes the `"$(cat "$SKILL_PATH")"` positional, (c) `CLAUDE_SCHEDULED_TASK` defaults to 1 only if not already set (env propagates correctly from plist), (d) exit codes 2/3/4 are distinct (skill-name-missing/skill-md-missing/claude-not-on-path).
- Verified `install-launchd-routines.sh` diff: preflight inside install branch only ✓, `gh || true` ✓, validator sourced + invoked ✓, render sentinel-check before mv ✓, plist env adds CLAUDE_SCHEDULED_TASK ✓, --uninstall removes rendered wrapper ✓.
- Confirmed `scripts/lib/launchd-routines-e2e-test.sh` test count is 15 (T1-T8, T9, T10-T14, T4a) and matches the codex-approved R7 scope + the R-impl-1 T4a addition.
- Spot-checked T8 argv assertion: argc=9 (was 8 before `--` was added), argv[8]=`--`, argv[9] starts with SKILL.md frontmatter `---`. Coherent.

### FU follow-through

#226 transcript-walker bug is correctly filed as separate scope. The mining routine producing 0 candidates today is the SKILL's problem (root-owned `.jsonl` files + walker fragility), not the wrapper-mechanism's. Wrapper-fix stability is independent.

### Disposition

**Comment** (not approval). Author signs off via the GitHub UI for the `review-required` gate; this comment satisfies the Rule-17 substantive-bot-review requirement but does not satisfy `review-required` itself.

Wait for #221 to land, then rebase this onto main. The duplicate-commit detection during `git rebase main` should drop the in-progress #221 commits automatically once they exist as their canonical hashes on main.

lantiscooperdev (approved):

PR #335, opened 2026-05-23
fix(classifier): em-rebuild-index → read_only (metadata sync)

lantisprime:
## Summary

- Drops the `shared_write` special-case for `em-rebuild-index.mjs` in `hooks/lib/command-classifier.sh:1480-1487`; it now classifies `read_only` like its sibling em-* read scripts.
- Rationale: rebuild is deterministic metadata sync derived from episode files (idempotent, atomic-rename, no partial-corruption window). Same gate-class as `em-search` (which also writes — access_count). The prior `shared_write` label was false-positive-blocking checkpoint-gate when `.checkpoint-required` is armed, forcing manual checkpoint-marker writes for metadata sync that is not Rule-18-scope implementation work.
- Adds regression test T16b in `tests/test-command-classifier.sh`. Full suite: 341/341 pass.

## Test plan

- [x] `bash tests/test-command-classifier.sh` → 341 pass, 0 fail
- [ ] In a session where `.checkpoint-required` is armed and `.pre-checkpoint-done` is missing, `node scripts/em-rebuild-index.mjs --scope all` passes the gate directly (no manual marker write needed).
- [ ] No regressions to em-store / em-revise / em-prune classification (still `shared_write`).

## Context

Same false-positive class as the open observation in episode `20260521-234705` (validate-*.mjs blocked as `shared_write` by checkpoint-gate). User has hit this every session that arms `.checkpoint-required`.

### DIFF
--- a/hooks/lib/command-classifier.sh
+++ b/hooks/lib/command-classifier.sh
@@ -1478,11 +1478,11 @@ _classify_segment() {
       script_base="$(basename "$script" 2>/dev/null)"
       case "$script_base" in
         em-search.mjs|em-list.mjs|em-watch-codex.mjs|em-pattern-health.mjs|em-check-stale.mjs|em-rebuild-index.mjs)
-          # em-rebuild-index touches index.jsonl — treat as shared_write
-          if [ "$script_base" = "em-rebuild-index.mjs" ]; then
-            printf '%s\t\t%s\n' "shared_write" "interpreter_em_rebuild"
-            return 0
-          fi
+          # em-rebuild-index writes index.jsonl but the operation is metadata
+          # sync derived deterministically from episode files (idempotent,
+          # atomic-rename, no partial-corruption window). Same gate-class as
+          # em-search (which also writes — access_count). Treating as
+          # read_only so the gate stops false-positive-blocking metadata sync.
           printf '%s\t\t%s\n' "read_only" "interpreter_em_read"
           return 0
           ;;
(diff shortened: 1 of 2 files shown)


lantiscooperdev (commented):
## Adversarial review — PR #335

Walked the diff, the gate flow, and downstream call sites.

### Findings

**ACCEPT — change is minimal and the rationale holds.**

- Diff is 5 lines of substantive code (delete the inner `if`) + a new test line. No collateral.
- Sibling parity is real: `em-search` writes `access_count` under `--track` and still classifies `read_only`; treating `em-rebuild-index` differently was inconsistent.
- Idempotence + atomic-rename are verifiable in `scripts/em-rebuild-index.mjs` (writes a temp file then renames; no partial-state window).
- `grep -rn "interpreter_em_rebuild"` returns zero hits — no downstream consumer pinned to the old reason string.

### Risk surface considered

1. **Push-gate bypass?** No. Push-gate keys on `push_or_pr_create`, not `shared_write` vs `read_only`. Reclassifying em-rebuild-index doesn't loosen push enforcement.
2. **Pre-checkpoint bypass for actual implementation work disguised as rebuild?** The script's argv is fixed; the classifier matches the basename `em-rebuild-index.mjs` and the script_digest is captured by the script-digest cache key. An attacker who can swap the script contents already controls the host.
3. **Threat model for index.jsonl corruption?** Same as before — atomic-rename is the existing defense; classifier label doesn't affect it.
4. **Test coverage?** T16b covers the canonical `--scope all` form. Other flag combinations (`--scope local`, etc.) classify identically because the dispatch happens on `script_base` only, not on argv.

### Verdict: ACCEPT

No blockers, no HOLDs. Squash-merge when ready.

lantiscooperdev (approved):

Issue #351, opened 2026-05-25
Planning-passive F1 residual: pure-Bash implementation bypasses pre-checkpoint (PR-B agent-classifier dependency)


lantisprime:
## Summary

The checkpoint-gate **planning-passive redesign** (branch `fix/checkpoint-planning-passive`) removes Bash from the pre-checkpoint gate so that planning / discovery / exploration / code-review never block and never arm a marker. The pre-checkpoint requirement now lazily arms at the **first repo-source `Edit`/`Write`/`MultiEdit`/`NotebookEdit`** (the implementation boundary).

**F1 residual (user-accepted):** because Bash is ungated by the pre-checkpoint gate, a *pure-Bash implementation* can mutate repo source without ever arming `.checkpoint-required`, bypassing the Rule 18 pre-implementation checkpoint.

This issue tracks the dependency on the **rank-10 PR-B agent-classifier** for clean closure.

## 5-field DEFER justification

| Field | Detail |
|---|---|
| **1. Run the scenario** | `echo 'x' > src/foo.mjs` / `sed -i ... file` / `node writer.mjs` from a fresh session arms nothing → no pre-checkpoint block. Reproduced via `tests/test-checkpoint-gate.sh` tests 64–67b / 81–89 (chained `rm`/heredoc shared_write now ALLOWED) and the `PP-1..PP-3` planning-passive section (shared_write Bash allowed, nothing armed). |
| **2. Spec check** | RFC-002 Phase 3b requires the pre-checkpoint to gate *write tools*; it does not mandate gating arbitrary Bash. No spec line requires Bash pre-gating, so this is a deliberate scope decision, not a spec violation. |
| **3. History check** | Gating all shared_write Bash is exactly what produced the recurring planning-time friction (read-only `node`/inspection commands mis-blocked) documented across the checkpoint-gate recurrence history and `feedback_preclassify_novel_commands.md`. |
| **4. Same-class check** | Class = "repo mutation that should require a pre-checkpoint." Members: Edit/Write/MultiEdit/NotebookEdit (GATED — lazy-arm) and Bash mutation (UNGATED — this residual). The Edit-class members are fully covered; only the Bash-mutation sibling is deferred. |
[...]

Issue #333, opened 2026-05-23
Wire deny-with-hint UX for marker-miss on interpreter commands (PR #331 deferred design item)


lantisprime:
## Context

PR #331 (`feat(classifier): replace direct-API Tier 3 with agent-self-classify marker`, merged `231045d`) shipped the marker-cache + helper but did NOT ship the **active deny-with-hint** mechanism the design plans (design R1–R3) described.

Current behavior on marker miss:

1. Bash invocation hits `hooks/lib/command-classifier.sh:~L1474` (interpreter case, non-allowlisted script).
2. `llm_classify_command` in `hooks/lib/llm-classifier.sh` reads `<project>/.checkpoints/classify/<sha>.json`.
3. Marker absent / stale / tampered → wrapper returns `rc=1` (no decision).
4. Caller falls to Tier 1 default: `shared_write / interpreter_other` (conservative).
5. Downstream `pre-checkpoint-gate` blocks the Bash IF no checkpoint marker exists, with its own reason — but that reason doesn't reference the agent-self-classify path.

So the agent doesn't get a clear actionable nudge "write a classifier marker to override this." It just sees a generic checkpoint-gate block. To actually populate the cache, the agent has to either:

- Already know about `scripts/classifier-marker.mjs --write` and invoke it proactively before the Bash call (out-of-band discipline)
- See the conservative `shared_write` classification, reason "no, this is actually `read_only`," and manually invoke the helper

Neither is enforced or hinted at the hook layer.

## What the design plans described (and what didn't ship)

From the design-R3 ACCEPT plan (codex episode `20260523-093630-…-386b`):

> Hook deny reason format (the contract with the agent):
>
> ```
> Classify this Bash command before retrying.
> Write the verdict to .checkpoints/classify/<sha>.json:
>
>   node scripts/classifier-marker.mjs --write \
>     --project-root <abs> \
>     --caller-cwd <abs> \
>     --command "<text>" \
>     --label <read_only|marker_write|shared_write|push_or_pr_create|unsafe_complex> \
>     --confidence <0..1> \
>     --reason "<one-line>"
>
> Then retry the Bash call.
> ```

This text was specified as the deny-reason an agent would see. The implementation **didn't emit it** — `llm_classify_command` just returns `rc=1` and the caller falls to the conservative default.

## Why it didn't ship

[...]

PR #352, opened 2026-05-25
fix(checkpoint-gate): planning-passive redesign — lazy-arm at first repo write, no session-start arming

lantisprime:
## Summary

Eliminates checkpoint friction during planning. The pre-checkpoint gate no longer arms at SessionStart, and **Bash is removed from the pre-checkpoint gate** — planning / discovery / exploration / code-review never block and never create a marker. The pre-checkpoint requirement now **lazily arms on the first repo-source `Edit`/`Write`/`MultiEdit`/`NotebookEdit`** (the implementation boundary). The push-gate + post-checkpoint + stop-gate lifecycle stay fully enforced.

## What changed
- **`hooks/checkpoint-gate.sh`**
  - Pre-gate is Edit/Write-only with lazy-arm, guarded by a concrete in-repo `FILE_PATH` — the empty-path conservative block no longer leaks `.checkpoint-required` into the hook-cwd repo (caller-leak regression).
  - Marker-write branch is self-contained: a premature `.post-checkpoint-done` write blocks directly via `_block_pre` (was relying on the removed Bash pre-gate).
  - **Env-prefix wrapper-escape block** on classifier-marker invocations (`REASON=classifier_marker_env_override`) — rejects the *form* (`BYPASS=1 …`, `NODE_ENV=…`), not a var-name list.
- **`scripts/em-recall.mjs`** — removed the dead `armCheckpointMarkerForSession`; a recent bp-001 violation now emits the `__BP1_ADVISORY__` stderr signal instead of arming.
- **`hooks/em-recall-sessionstart.sh`** — surfaces the advisory via SessionStart context.
- **Tests** — +16 planning-passive `PP-*` tests (nothing-armed flow, lazy-arm, off-repo allow, multisession, multiproject); ~34 gate tests + 5 em-recall/rfc002/migration suites updated to the advisory + ungated-Bash model.

## F1 residual (accepted)
A pure-Bash implementation (`sed -i`, `cat > file`, `node writer.mjs`) can mutate repo source without arming the pre-checkpoint — a deliberate trade to remove planning-time friction. Clean closure depends on the rank-10 PR-B agent-classifier. Tracked in #351.

## Testing
- `tests/test-checkpoint-gate.sh` → 236/0
- Full CI sweep green: command-classifier 361/0, em-recall suites, rfc002-phase3 29/0, checkpoints-migration 7/0, bp001-per-project-arming 8/0, repo-root-worktree 12/0
[...]

PR #353, opened 2026-05-25
feat(agent-classifier): rename + --command-file(#349) + env aliases (F1 #351 + deny-hint #333 → PR-B2)

### LINKED ISSUE #349: _validate_classifier_marker_helper awk extraction breaks on shell-quoted paths with spaces
Surfaced by claude-subagent PR-level review of PR-A (commit `97e26a0`). Reviewer finding PR-A-PR-LEVEL-NIT-2.

## Background

`hooks/checkpoint-gate.sh:687-694` `_validate_classifier_marker_helper` extracts the script path via awk:

```bash
script_path="$(printf '%s' "$cmd" | awk '{
  for (i=1; i<=NF; i++) {
    if ($i == "node" || $i == "python" || $i == "python3" || $i == "ruby" || $i == "perl") {
      if (i+1 <= NF) print $(i+1)
      exit
    }
  }
}')"
```

`awk` splits on whitespace. A shell-quoted path like:

```
node "/Users/some user/.episodic-memory/scripts/classifier-marker.mjs" --write ...
```

extracts `"/Users/some` (leading quote, truncated at the space inside the quoted string). Then `_canonicalize_possibly_nonexistent "/Users/some"` returns a non-matching path, validator returns 1, carve-out doesn't fire, pre-gate blocks the legitimate canonical invocation.

## Asymmetric handling

Upstream classifier at `command-classifier.sh:1638` uses `${TOKS[$((idx+1))]:-}` where `_tokenize` properly handles shell quoting. So:

- Classifier sees the correctly-tokenized path → labels `marker_write/interpreter_classifier_marker`
- Validator's awk extraction returns the broken truncated path → carve-out refuses
- Pre-gate blocks the canonical helper invocation

Fail-closed direction (no security impact), but creates UX friction for:
- macOS users whose `$HOME` contains spaces (common; e.g. `/Users/Some User/`)
- Repos checked out under paths with spaces (less common but possible)

## Impact

Currently low — no user has reported this. But the validator becomes more important once PR-B ships the deny-with-hint flow (the canonical helper invocation is the *resolution* path for blocked novel commands; if the validator fail-closed-blocks the canonical helper itself, the whole rank-10 UX breaks for affected users).

## Two acceptable fixes

**(a) Replace awk with `_tokenize`-driven extraction.** Source `command-classifier.sh`'s `_tokenize` (or extract a smaller shared tokenizer to `hooks/lib/token-extract.sh`), use it for both classifier dispatch AND validator extraction. Single source of truth for "what's the script path token." Larger refactor; needs care to avoid sourcing the whole classifier into checkpoint-gate.sh.

**(b) Document the limitation in `_validate_classifier_marker_helper`'s comment block.** Adjacent to the existing `--inspect-brk` / `.cjs` extension caveats at `hooks/checkpoint-gate.sh:678-685`. Doesn't fix the bug, but stops it being a hidden gotcha. Smaller; doesn't unblock the UX edge case.

Recommend (a) as it removes the bug class. (b) is a fallback if (a) is too much churn.

## Test plan

Regression test: a test fixture under `$HOME/Some User/` or equivalent space-containing path; assert validator returns 0 (canonical match) when invoked via a quoted-path command.

## References

- PR-A: `fix/classifier-authority-root-and-bootstrap-carveout` (commit `97e26a0`)
- Claude-subagent PR-level review: episode `20260524-141649-reply-claude-subagent-to-20260524-141133-f9c2`
- Reviewer finding: PR-A-PR-LEVEL-NIT-2 (MINOR per reviewer; UX issue, not security)

## Priority

P3. Not blocking PR-A. Bundle with rank-10 PR-B's deny-with-hint work since the UX impact gets larger there.

lantisprime:
## Summary

Renames the `llm-classifier` subsystem to `agent-classifier`, adds `--command-file` + fixes the #349 quoted-path extraction, and renames the env surface with backward-compat aliases.

**Closes #349.** **#333 (deny-hint) and #351 (F1 pre-checkpoint closure) are deferred to PR-B2** — see "Scope change" below.

## What changed

1. **Rename llm-classifier → agent-classifier.** `git mv hooks/lib/llm-classifier.sh → agent-classifier.sh`, `scripts/llm-classifier-dispatch.mjs → agent-classifier-dispatch.mjs`; `llm_classify_command → agent_classify_command`, sentinel/internals renamed; sourcing updated. Source-tags `llm-*-autopersist → agent-*` (old kept as backward-compat in `VALID_SOURCE_TAGS`). **Install-time orphan sweep** (`install.mjs`) deletes old-named installed copies — gated on the new files being current this run (withheld if `command-classifier.sh` is `skipped-divergent`).
2. **`--command-file` + #349.** `classifier-marker.mjs` accepts `--command-file` (mutually exclusive with `--command`, 64 KiB cap, verbatim read). #349: `_validate_classifier_marker_helper` now extracts the script path via the shared de-quoting `_tokenize` tokenizer instead of whitespace-split `awk`, so quoted paths with spaces work.
3. **Env rename + aliases.** `AGENT_CLASSIFIER_*` primary; `LLM_CLASSIFIER_*` still works as a backward-compat alias (new wins if both set; deprecation note on stderr when only the old name is seen). `AGENT_CLASSIFIER_DISPATCH_PATH` alias too. Docs updated.

## Scope change (was: also F1 #351 + deny-hint #333)

This PR originally bundled the F1 pure-Bash pre-checkpoint closure (#351) and its deny-hint UX (#333). The Codex PR-level review demonstrated that the **enumerated content-write arm is unboundedly leaky** — each round surfaced more named-target writers (tee/cp/mv/dd → install/truncate/`:` → rsync/curl/wget → …), because the classifier's `default_write` bucket conflates *content-writers* with *process-mutations* and there's no syntactic signal to separate them.

The right fix is an **agent-classifier `nonsrc_write` verdict** — the agent (which actually knows what `rsync a.txt scripts/x` does) classifies the repo-source-write question, eliminating enumeration. That's a marker-schema increment deserving its own plan + review cycle. Per maintainer decision, **F1 (#351) + the coupled deny-hint (#333) move to PR-B2**; this PR ships the clean, Codex-accepted rename + `--command-file` + env-alias parts. (The F1 work is preserved on `fix/agent-classifier-pr-b2-f1-snapshot`.)

## Review

- **Plan-tier (Codex):** R1 HOLD → R2 ACCEPT-with-FU.
- **Code-tier (negative-scenario-reviewer):** found the F1 `tee/cp/mv/dd` gap (now deferred with F1) + deny-hint framing (deferred); rename/`--command-file`/orphan-sweep verified clean.
- **PR-level (Codex):** R1+R2 HOLD on F1 class-completeness → resolved by the split; Codex explicitly confirmed "rename integration, `--command-file`, orphan sweep, `.review-store` carve-out look coherent."

## Tests

All green: command-classifier **361**, checkpoint-gate **237** (incl. a #349 spaced-path test; planning-passive review-freedom intact), agent-classifier **40** (incl. env-alias matrix), classifier-marker **21** (incl. `--command-file`), install-hooks **81** (incl. orphan sweep + corrected manifest count). `test-checkpoint-gate.sh` + `test-install-hooks.sh` were in no CI workflow — now wired in.


lantiscooperdev (commented):
## Bot review — PR #353 (agent-classifier / PR-B)

Substantive pre-merge review (bot account; **user approval still required** in the Reviews tab per Rule 17).

### Verified ✅
- **Rename completeness:** no shipped code references `llm_classify_command`/`__llm_classifier`/`llm-classifier.sh` except the intentional `RENAMED_REMOVED` sweep list, env aliases, and one historical-note comment. `git mv` preserves blame on the two renamed files.
- **Orphan sweep is safe:** gated on the new files being current this run; withholds (with a clear message) when `command-classifier.sh` is `skipped-divergent` — proven by `test-install-hooks.sh` T19a–f. Explicit basename list, never glob-delete.
- **F1 narrow closure is consistent:** `cmd_content_write` (tee/cp/mv/dd) + redirect reasons arm; `interpreter_other` / process-mutations / read_only stay free. Verified end-to-end on the **deployed** gate (review-dispatch + `git commit` + `ls` all free; `echo>repo`, `cp>repo`, `tee>repo` block).
- **`.review-store/` carve-out** is canonical-checked — a symlink escaping to repo source still arms (test F1-11).
- **Deny-hint** single-quote escaping round-trips under `eval` for adversarial inputs (embedded quotes, `$()`, backticks, newlines); checkpoint-first framing with anti-mislabel warning.
- **CI gap closed:** `test-checkpoint-gate.sh` + `test-install-hooks.sh` were in no workflow; now wired in. The latter immediately caught a stale `T1b8` count (10→12).

### Decisions worth a human eye before merge 👀
1. **F1 was narrowed** from the originally-chosen "re-gate all mutating Bash" to "content-writes only," because the broad form re-blocked `node second-opinion.mjs` review dispatch — undoing a PR #352 win. This is the right call IMO, but it means **documented residuals remain**: `node writer.mjs` (interpreter_other) and `sed -i` (classifier labels it `read_only`) bypass the pre-checkpoint. The push-gate still catches the eventual push. Confirm you're comfortable with that residual surface.
2. **External-target over-arm:** `cat > /tmp/x` / `tee /tmp/x` arm the pre-checkpoint (no write-target discrimination in the classifier). The deny-hint provides a one-time classify-read_only escape. Mild friction on Bash-redirect-to-/tmp during planning; rare in practice (scratch usually goes through the Write tool, which is off-repo).

### Notes (non-blocking, not fixed here)
- `--command-file` fallback in the deny-hint suggests a `<repo>/.checkpoints/classify/pending-*.cmd` scratch; writing that via a Bash redirect would itself arm. The primary (inline `--command`) path is unaffected; the agent would create the file via the Write tool. Flagging for the Codex PR-level pass.
- `classifier-config-loader.mjs` main-guard (`import.meta.url === file://argv[1]`) is symlink-fragile (surfaced only because the E2E ran the deployed copy under macOS `/tmp`). Pre-existing; production `HOME=~` is unaffected.

Recommending **approve** pending the Codex PR-level review now in flight.

lantiscooperdev (commented):
## Bot review (updated for reduced scope) — PR #353

Superseding my earlier review: this PR was **scoped down**. F1 (#351) + deny-hint (#333) are deferred to PR-B2 after the Codex PR-level review showed the enumerated content-write arm is unboundedly leaky (the classifier can't separate content-writers from process-mutations within `default_write`). This PR is now the **Codex-accepted subset**: rename + `--command-file`(#349) + env aliases.

### Verified on the reduced diff ✅
- **Clean revert:** no dangling refs to `_block_pre_with_hint` / `_bash_reason_arms_pre_checkpoint` / `agent_classifier_deny_hint` / `cmd_content_write` / `agent-classifier-deny-reason.sh` in shipped code (only the intentional PR-B2 forward-reference in the F1 RESIDUAL header). `bash -n` + `node --check` clean.
- **Planning-passive restored:** Bash is ungated again (PP-3 review-dispatch-free passes); the F1 RESIDUAL header is back to the #352 text, now pointing at the PR-B2 `nonsrc_write` plan.
- **Rename + orphan sweep:** install T19a–f (sweep + divergent-skip) green; manifest count corrected 12→11 after the deny-reason lib removal.
- **#349** `_tokenize` de-quote extraction retained + its spaced-path gate test.
- **Env aliases** new-wins / old-alias+deprecation / both-set matrix green.
- All suites green: command-classifier 361, checkpoint-gate 237, agent-classifier 40, classifier-marker 21, install-hooks 81.

### Tracking
- #333 reopened, #351 annotated with the PR-B2 `nonsrc_write` design. F1 work preserved on `fix/agent-classifier-pr-b2-f1-snapshot`.

No blockers on the reduced scope. Recommending **approve** (user approval required per Rule 17).

lantiscooperdev (approved):

PR #356, opened 2026-05-28
feat(second-opinion): add opencode (DeepSeek v4-pro) provider + surface spawn errors

lantisprime:
## Summary
Adds an **OpenCode** provider (model `deepseek/deepseek-v4-pro`, overridable via `OPENCODE_MODEL`) to the second-opinion review harness, and fixes a cross-provider follow-up (FU-001) surfaced during the live review.

## Changes
- **New provider** `scripts/second-opinion/providers/opencode.mjs` — `available()` + `dispatch()` (`opencode run -m <model>`, `shell:false`, explicit cwd, `stdin:ignore`).
- **Registry** `providers/index.json` — `cli_match: ^opencode\s+run\b`, scoped to the review subcommand so the gate blocks only `opencode run`, not the interactive TUI or other subcommands.
- **Preamble** default + new `fragments/opencode-ladder-v1.md`.
- **FU-001 fix (all 4 CLI providers)** — `dispatch()` now surfaces spawn failures (ENOENT / bad cwd) via an `error` field instead of discarding `result.error`. `stub` is exempt (in-process).
- **Docs** — README + USER_MANUAL provider lists; new maintainer runbook `hooks/runbooks/second-opinion-add-provider.md`.

## available() gotcha fixed
OpenCode prints `--help` to **stderr**; the initial stdout-only probe returned `cli-help-signature-mismatch` at install. `available()` now reads both streams and matches the signature against their concatenation.

## Tests (all green)
- providers 24/0 (incl. 4 FU-001 cases + opencode contract)
- dispatch 9/0 · gate 23/0 (cli_match scoping: `opencode run` gated, bare `opencode` allowed)
- preamble 44/0 · storage 11/0 (episodic request/reply round-trip + `em-search` readback)
- install-snapshot 14/0 · install-e2e 6/0

## Live E2E
`opencode` → DeepSeek v4-pro returned **ACCEPT** on both `--storage files` and `--storage episodic`. The episodic request/reply were written as local episodes and read back via `em-search --tag reply-to-<id>` — the same path codex replies use.

## Notes
- Second-opinion (codex) review on the plan was waived per request.
- The live opencode review of `opencode.mjs` served as a de-facto code review (ACCEPT + FU-001, now fixed inline across all providers).

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,7 +418,7 @@ node scripts/second-opinion.mjs request \
   --consensus --max-rounds 5 --rebuttal-cb scripts/my-rebuttal.mjs
 ```
 
-Providers: `codex`, `claude-subagent`, `gemini`, `stub` (testing). Storage backends: `files` (`.review-store/`) or `episodic` (uses em-store; default for the cross-tool message bus). Preambles: per-provider defaults at `scripts/second-opinion/preambles/`, overridable via `--preamble <id>` or `<project>/.review-store/preambles/<provider>.md`.
+Providers: `codex`, `claude-subagent`, `gemini`, `opencode` (OpenCode CLI; DeepSeek v4-pro by default, overridable via `OPENCODE_MODEL`), `stub` (testing). Storage backends: `files` (`.review-store/`) or `episodic` (uses em-store; default for the cross-tool message bus). Preambles: per-provider defaults at `scripts/second-opinion/preambles/`, overridable via `--preamble <id>` or `<project>/.review-store/preambles/<provider>.md`.
 
 Bootstrap (writes the install snapshot consumed by validators + the Claude Code PreToolUse gate hook):
 
--- a/docs/USER_MANUAL.md
+++ b/docs/USER_MANUAL.md
@@ -334,7 +334,7 @@ AI:   I see from a previous session that you chose Tailwind CSS
 
 **How it works:** Both tools read from and write to `~/.episodic-memory/`. The instruction files are different per tool, but the data is shared.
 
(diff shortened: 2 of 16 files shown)


lantiscooperdev (commented):
**Automated review (bot account) — PR #356: opencode provider + FU-001**

Reviewed the diff and adjacent provider code. CI `validate` green; live E2E confirms `opencode` → DeepSeek v4-pro returns ACCEPT on both `files` and `episodic` backends.

### Strengths
- **cli_match scoping** — `^opencode\s+run\b` (not the bare binary) gates only the review invocation. The gate test asserts both directions: `opencode run …` blocked, bare `opencode` / `opencode models` allowed. This is the right call.
- **available() stderr handling** — OpenCode prints `--help` to stderr; reading both streams is correct (and an improvement over codex/gemini, which read stdout only). The unit tests can't catch this — the install `Registered`/`Skip` line is the real signal.
- **FU-001 fix is uniform** — `error` field added across all 4 CLI providers (stub correctly exempt), with a deterministic bad-cwd regression test that doesn't depend on which CLIs are installed.
- **e2e same-class catch** — `makeAllProvidersUnavailable` updated to include opencode; required for Case 4 to fire on hosts where opencode is present.
- **Episodic round-trip** now has a hermetic regression test (write request+reply episodes + `em-search` readback).

### Findings (all informational — non-blocking)
1. **[info] Gate anchoring.** `^opencode\s+run\b` won't match an env-prefixed call (`FOO=bar opencode run …`). Same pre-existing limitation as codex's `^codex\s+exec\b`; not introduced by this PR. Noting for the shared gate-bypass class.
2. **[info] No explicit read-only guard** on the agentic `opencode run` — relies on the ladder fragment's "do not edit files" instruction. Acceptable for v1 (a non-TTY run won't get interactive tool approval); revisit if OpenCode later auto-approves tools.
3. **[nit] available() signature coupling.** The `opencode run` signature is tied to help wording; a future help reword would break detection. Low risk, consistent with sibling providers.

### Verdict: ACCEPT
No blockers. CI green, live E2E ACCEPT, 8 test suites cover the contract, gate scoping, FU-001, and the episodic round-trip. Findings 1–3 are informational and don't require changes before merge. User approval in the UI gates the merge (Rule 17).

lantiscooperdev (approved):

PR #313, opened 2026-05-17
feat(rfc-004): slice 2e — BP-1 deadline-firing path (check-deadlines + A2 + state-lock + crash-classify)

lantisprime:
## Summary

RFC-004 M2 slice 2e — BP-1 auto-pilot **deadline-firing path** (A2 auto-approve on TTL). Lands the missing piece between slice 2d (W: record-awaiting-approval; R: H1 marker reader + confirm-approval) and slice 2f (A1 retry-tree): the **`check-deadlines` orchestrator subcommand** that walks the firings queue and dispatches A2 transitions per `<canonical_project_root>` with full audit-trail closure.

**Scope (C1-C6):**
- **C1** `tryAcquireRunStateLock` non-blocking primitive (`bd4739f`) — RFC §1196-1212
- **C2** `bp1-deadlines.mjs` pure A1/A2 helpers (`c8313d1`)
- **C3** `bp1-crash-classify.mjs` 17-row state walker (`4fdbb2c`)
- **C4.3** canonicalize +5 episode types (`276c4f6`) — contract v3.16→v3.17
- **C4.1** `bp1-state-lock.mjs` O_EXCL + signed claim episode + 60s TTL stale-break (`5c401f2`)
- **C4.2** `bp1-request-claim.mjs` forward-ready wrapper (`585d58c`)
- **C4.4** `parseArgs` M5 hardening — reject unknown flags / unexpected positionals / missing values (`8f6d178`)
- **C4.5** `check-deadlines` subcommand + `fireA2` path (`b8d5645`) — RFC §1194-1233
- **Review fixes** (`9ace44c`) — 4 inline P2 fixes from `negative-scenario-reviewer` HOLD

## Negative-scenario-reviewer dispositions

Reply episode `20260517-122541-c4-review-hold-1-p1-cwd-binding-git-topl-fae8`. Verdict: HOLD with 1 P1 + 4 P2 + 3 P3.

| Finding | Verdict | Action |
|---|---|---|
| C-P1.1 git-toplevel "hijacks" authority root | **REJECT** | RFC §104 mandates `git rev-parse --show-toplevel + realpath`; CD10 codifies design (subdir-of-git → toplevel). Same resolver pattern at `init-run` L2104 + `confirm-approval` L2470. |
| A-P2.1 orphan lockfile on emit failure | **ACCEPT** | inline try/catch unlink + SL11 regression |
[...]

PR #363, opened 2026-05-28
fix(checkpoint-gate): allow .post-checkpoint-done write when PRE_DONE non-empty (#362)

### LINKED ISSUE #362: bug: checkpoint-gate.sh refuses .post-checkpoint-done write when .pre-checkpoint-done exists, emits misleading 'write pre-checkpoint' message
## Symptom

After writing a non-empty `.pre-checkpoint-done.<sid>`, attempting to write `.post-checkpoint-done.<sid>` via Write/Edit is rejected with:

> "Checkpoint required. Write the Rule 18 **pre-implementation** checkpoint block to `.checkpoints/.pre-checkpoint-done.<sid>` (must be non-empty) before write tools are unblocked. Hook: checkpoint-gate.sh."

— even though that exact pre-checkpoint marker is already on disk and non-empty.

Empirically reproduced 2026-05-28 session `ca3b5e2f-4c80-4c26-9b02-21a247a65770` while working on PR #361 (the #358 regression-test PR). `ls -la .checkpoints/` showed `.pre-checkpoint-done.ca3b5e2f-...` at 722 bytes; the next `Write` to `.post-checkpoint-done.ca3b5e2f-...` was still blocked with the pre-checkpoint message.

## Root cause

`hooks/checkpoint-gate.sh:1185-1196`:

```bash
.post-checkpoint-done|.post-checkpoint-done.*)
  if checkpoint_marker_exists_for_session .post-checkpoint-required "$MY_SID"; then
    # POST_REQ armed → post-checkpoint phase; allow the write
    exit 0
  fi
  # POST_REQ NOT armed → premature marker_write
  _block_pre
  ;;
```

The gate decides 'can I write `.post-checkpoint-done`?' **purely** on whether `.post-checkpoint-required` is armed. It never reads `.pre-checkpoint-done`.

`.post-checkpoint-required` is armed **only by the push-gate** (line 1485, comment: 'push self-arms .post-checkpoint-required'). So the gate's lifecycle expects:

1. `git push` attempt
2. push-gate self-arms `.post-checkpoint-required` and blocks with 'post-checkpoint required'
3. agent writes `.post-checkpoint-done`
4. agent retries push → succeeds

But the natural agent flow is *finish work → write post-checkpoint → push*, which deadlocks at step 1 (can't write post-checkpoint until push, but push needs post-checkpoint already written). The work-around is to attempt the push, get blocked, then write the marker — but that's a 'deadlock and rescue' pattern, not a workflow.

Secondary problem: when the gate refuses the post-checkpoint write, it calls `_block_pre` (line 1196), which emits the **pre-implementation** error message pointing at `.pre-checkpoint-done.<sid>` — even though that file is already non-empty. The message is misleading: the actual missing prerequisite is `.post-checkpoint-required` being armed (an internal gate state, not something the agent writes directly).

## Empirical artifact

This session's `.checkpoints/` state at the time of the block:
```
-rw-r--r-- ... 722 28 May 13:16 .pre-checkpoint-done.ca3b5e2f-4c80-4c26-9b02-21a247a65770
-rw-r--r-- ... 4267 28 May 13:14 .preflight-done.ca3b5e2f-4c80-4c26-9b02-21a247a65770
(no .post-checkpoint-required, no .post-checkpoint-done)
```
Subsequent `Write` to `.post-checkpoint-done.<sid>` returned the pre-checkpoint block message.

Workflow followed to escape:
1. `Write .post-checkpoint-done.<sid>` → blocked with pre-checkpoint message (misleading).
2. `Write .pre-checkpoint-done.<sid>` (overwriting the one already present) → succeeded.
3. `Write .post-checkpoint-done.<sid>` → STILL blocked with same message.
4. `git push` → blocked with **post**-checkpoint message (correct), push-gate self-armed `.post-checkpoint-required`.
5. `Write .post-checkpoint-done.<sid>` → now succeeded.
6. `git push` → succeeded.

## Suggested fix

Two-part:

1. **Allow `.post-checkpoint-done` write when `.pre-checkpoint-done` is non-empty for the session**, regardless of `.post-checkpoint-required` state. The agent declaring 'impl complete' after a valid 'impl start' is the natural workflow. Push-time enforcement (line 1525) already independently checks `.post-checkpoint-done` is non-empty, so this doesn't weaken push gating.

   At line 1186:
   ```bash
   if checkpoint_marker_exists_for_session .post-checkpoint-required "$MY_SID" \
      || checkpoint_marker_nonempty_for_session .pre-checkpoint-done "$MY_SID"; then
     exit 0
   fi
   ```

2. **Fix the misleading message.** When refusing post-checkpoint-done write, the agent shouldn't be told to write pre-checkpoint (which may already exist). Emit either:
   - 'Post-checkpoint phase not entered — complete a write that arms `.post-checkpoint-required` (e.g. `git push`) first.' (current behavior, just clearer)
   - Or — if fix #1 is adopted — only block when pre-checkpoint is ALSO missing, in which case `_block_pre` is correct.

## Tests affected

`tests/test-checkpoint-gate.sh` cases 17/49/50/56 currently assert that premature post-checkpoint-done writes are blocked (per comment at line 1195). Under fix #1 those tests would need their setup to also lack `.pre-checkpoint-done` (the 'truly premature' case). Then add new cases:
- `pre-checkpoint present + post-checkpoint write → ALLOWED` (regression for this bug).
- `no pre-checkpoint + no post-required + post-checkpoint write → BLOCKED` (existing semantics preserved).

## Severity / impact

Annoyance-level — work always completes via the workaround — but the misleading error message is a real correctness bug: the agent is told to write a marker it just wrote. New users / fresh agents could overwrite a valid pre-checkpoint marker chasing a phantom failure.

## Related

- Closed by #360 + #361: classifier marker-cache key drift (#358), DIFFERENT bug.
- This issue: `checkpoint-gate.sh` lifecycle ordering.

lantisprime:
## Summary

Fixes the workflow deadlock + misleading error message documented in #362.

The `hooks/checkpoint-gate.sh` post-checkpoint-done marker-write predicate previously keyed only on `.post-checkpoint-required` being armed. That marker is armed by either the allowed-write tail (`:1572` — requires an intervening Edit/Write/Bash between pre and post writes) or by push self-arm (`:1485`). The natural agent workflow `pre → finish → post → push` therefore deadlocked when no intervening work-write happened between pre and post — and the error message pointed at writing a marker that was already on disk and non-empty.

## Fix

Allow `.post-checkpoint-done` write when EITHER:
- `.post-checkpoint-required` armed for this session (pre-existing condition), OR
- `.pre-checkpoint-done` non-empty for this session (new).

Applied at both parallel branches:
- `hooks/checkpoint-gate.sh:1186` (Bash marker_write branch)
- `hooks/checkpoint-gate.sh:1330` (Edit/Write marker_write branch)

Top-of-file lifecycle comment at `:38-44` updated to reflect the new contract (codex ACCEPT-with-FU follow-up, folded inline).

## Push enforcement unchanged

- Push-gate at `:1485` still self-arms `.post-checkpoint-required` regardless of state.
- Push-gate at `:1525` still blocks push when `.post-checkpoint-done` is empty.

The fix only loosens the *intermediate* marker-write step; push-time gating remains the load-bearing enforcement. `_block_pre`'s "write pre-checkpoint" message is now accurate because that branch only fires when `.pre-checkpoint-done` is genuinely missing.

## Self-demonstration

This very session reproduced the bug while attempting to write the post-checkpoint block — the live (pre-fix, merged-main) hook refused the `Write` tool call because `.post-checkpoint-required` was not armed. The push-gate workaround (attempt push → push-gate self-arms POST_REQ + blocks → retry post-checkpoint write → push for real) was required, which is exactly what #362 documents.

## Tests

`bash tests/test-checkpoint-gate.sh` → **324 passed, 0 failed**. Pre-existing 307 + 17 new regression assertions:

- **I362-1..5** — directly exercise the predicate change (Bash + Edit allowed when PRE_DONE non-empty; cross-session BLOCKED; empty PRE_DONE still BLOCKED; push-gate independence preserved).
- **DL-EX-A.1..5** — multi-step workflow simulator walking the natural flow turn-by-turn: `pre → post (no intervening work) → push (block, self-arm POST_REQ) → push (allow, sweep convergence)`. Pre-fix this sequence deadlocked at step 2.
- **DL-EX-B.1..2** — Edit/Write tool parity for the workflow.
- **DL-EX-C** — anti-regression: truly-premature post-write (no pre-checkpoint anywhere) blocks with the accurate pre-checkpoint message.
- **DL-EX-D** — cross-session attack: session B's PRE_DONE does NOT unlock session A's post-write (per-session predicate).
- **DL-EX-E** — idempotent re-write preserved (post-done re-write after already non-empty stays allowed).

## Code review

Dispatched codex via `scripts/second-opinion.mjs request --provider codex --dispatch` against the staged diff with an adversarial prompt enumerating 8 attack axes (cross-session leak, zero-byte bypass, plan-pending invariant, wrong-root, stop-gate convergence, idempotent re-write, symlink classes, push-gate independence).

Reply episode `20260528-053742-reply-codex-to-20260528-053416-fix-check-70f4` returned **ACCEPT-with-FU**. The single FU (stale top-of-file lifecycle comment) is folded inline in this same PR per the inline-FU heuristic (sub-3-LOC same-surface fix).

Closes #362.

### DIFF
--- a/hooks/checkpoint-gate.sh
+++ b/hooks/checkpoint-gate.sh
@@ -702,6 +702,28 @@ _consume_plan_approved() {
   rm -f "$PRIMARY_DIR/$bn" "$LEGACY_DIR/$bn" 2>/dev/null || true
 }
 
+# Symlink-aware existence check for marker paths (codex round 3 P1, #364).
+# Mirrors marker-read predicates: a symlink at a marker path is NOT
+# considered present. Used by self-arm sites' pre-check to detect that a
+# fresh arm IS needed (the symlink will be replaced atomically by the
+# subsequent touch/write).
+_marker_path_has_real_file() {
+  [ ! -L "$1" ] && [ -e "$1" ]
+}
+
+# Symlink-safe touch: if a symlink exists at the path, unlink it first so
+# the touch creates a fresh regular file rather than following the symlink
+# to write outside the substrate. Best-effort under set -e (||true). Used
+# by the bash fallback sites that arm checkpoint markers when the
+# checkpoint-marker.mjs helper is not installed.
+_arm_marker_via_touch_safely() {
+  local p="$1"
+  if [ -L "$p" ]; then
+    rm -f "$p" 2>/dev/null || true
+  fi
+  touch "$p" 2>/dev/null || true
+}
+
 _arm_checkpoint_required_if_missing() {
   if checkpoint_marker_exists_for_session .checkpoint-required "$MY_SID"; then
     return 0
@@ -723,7 +745,10 @@ _arm_checkpoint_required_if_missing() {
       --action arm-if-missing \
       --root "$REPO_ROOT" >/dev/null 2>&1 || true
   else
-    touch "$(write_marker_path "$REPO_ROOT" "$(namespaced_marker_basename_for_session .checkpoint-required "$MY_SID")")" 2>/dev/null || true
+    # Symlink-safe (#364): unlink any planted symlink at the marker path
+    # before touch, otherwise touch follows the symlink and writes outside
+    # the substrate.
+    _arm_marker_via_touch_safely "$(write_marker_path "$REPO_ROOT" "$(namespaced_marker_basename_for_session .checkpoint-required "$MY_SID")")"
   fi
   # Transactional consume (codex PR review P1): the arm above is best-effort
   # (`|| true`) — an installed helper that exits non-zero would leave NO
@@ -1529,17 +1554,20 @@ if [ "$TOOL_NAME" = "Bash" ] && [ "$LABEL" = "push_or_pr_create" ]; then
       else
         # Fallback (helper not installed): compute existence BEFORE touch so the
         # created-now signal is race-free (§15-C3 — no read-after-arm window).
+        # Symlink-safe (#364): _marker_path_has_real_file returns false for a
+        # symlink, so this branch correctly enters the arm path; the helper
+        # touch then unlinks the symlink before writing.
         _push_req_path="$(write_marker_path "$REPO_ROOT" "$(namespaced_marker_basename_for_session .post-checkpoint-required "$MY_SID")")"
-        if [ ! -e "$_push_req_path" ]; then
-          touch "$_push_req_path" 2>/dev/null || true
+        if ! _marker_path_has_real_file "$_push_req_path"; then
+          _arm_marker_via_touch_safely "$_push_req_path"
           _push_self_arm_created=1
         fi
       fi
     else
       # Invalid/empty sid: legacy direct touch, existence-before-touch.
       _push_req_path="$(write_marker_path "$REPO_ROOT" .post-checkpoint-required)"
-      if [ ! -e "$_push_req_path" ]; then
-        touch "$_push_req_path" 2>/dev/null || true
+      if ! _marker_path_has_real_file "$_push_req_path"; then
+        _arm_marker_via_touch_safely "$_push_req_path"
         _push_self_arm_created=1
       fi
     fi
@@ -1613,14 +1641,14 @@ case "$TOOL_NAME" in
             --action arm-if-missing \
             --root "$REPO_ROOT" >/dev/null 2>&1 || true
         else
-          # Fallback: direct write when helper not installed.
-          touch "$(write_marker_path "$REPO_ROOT" "$(namespaced_marker_basename_for_session .post-checkpoint-required "$MY_SID")")" 2>/dev/null || true
+          # Fallback: direct write when helper not installed. Symlink-safe (#364).
+          _arm_marker_via_touch_safely "$(write_marker_path "$REPO_ROOT" "$(namespaced_marker_basename_for_session .post-checkpoint-required "$MY_SID")")"
         fi
       else
         # Invalid/empty sid: legacy direct touch (preserves pre-rank-2
         # behavior; gate inactive for own-session checks but legacy literal
-        # still works for old hooks).
-        touch "$(write_marker_path "$REPO_ROOT" .post-checkpoint-required)" 2>/dev/null || true
+        # still works for old hooks). Symlink-safe (#364).
+        _arm_marker_via_touch_safely "$(write_marker_path "$REPO_ROOT" .post-checkpoint-required)"
       fi
     fi
     ;;
--- a/hooks/checkpoint-gate.sh
+++ b/hooks/checkpoint-gate.sh
@@ -38,7 +38,10 @@ set -e
 #   - .pre-checkpoint-done write: requires .checkpoint-required exists
 #     (either root) AND .pre-checkpoint-done is missing/empty (both roots).
 #   - .post-checkpoint-done write: requires .post-checkpoint-required exists
-#     AND .post-checkpoint-done is missing/empty.
+#     OR .pre-checkpoint-done is non-empty for this session (#362 fix —
+#     allows the natural `pre → post → push` flow without an intervening
+#     work-write to arm POST_REQ). Push enforcement is independent: push-gate
+#     self-arms POST_REQ at :1485 and blocks at :1525 when POST_DONE is empty.
 # Cross-gate invariant (Codex ...3503 P1): checkpoint marker writes are
 # blocked while .plan-approval-pending exists (either root). Both gates
 # independently enforce.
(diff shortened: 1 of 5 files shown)


lantiscooperdev (commented):
## Bot review — fix(checkpoint-gate) #362

Reviewing the substrate-class gate change. Findings below; user approves in UI per Rule 17.

### Observations

1. **Predicate symmetry is clean.** Both Bash branch (`:1186`) and Edit/Write branch (`:1330`) use the identical disjunction `POST_REQ exists OR PRE_DONE non-empty for session`. No drift between the two enforcement sites — important because past PRs in this file (e.g. PR #313 R1-finding-1) have caught exactly this class of "same logic in two places, predicate diverges at one site."

2. **Session-scoping preserved.** `checkpoint_marker_nonempty_for_session` at `hooks/checkpoint-gate.sh:164-175` uses `[ -s "$path" ]` size check with session-suffixed basename lookup. Empty PRE_DONE does NOT unlock (test `I362-3`) and cross-session PRE_DONE does NOT unlock (test `I362-4` + `DL-EX-D`). Both attack surfaces are explicitly guarded.

3. **Push-time gate orthogonality.** The push-gate self-arm at `:1485` and the push-time POST_DONE-non-emptiness check at `:1525` are unchanged. The `B1-1..B1-5` push-cycle suite still passes (324/0), and new test `I362-5` explicitly asserts that push BLOCKS when PRE_DONE non-empty + POST_REQ armed but POST_DONE empty. So the load-bearing gate is preserved; the fix only loosens the intermediate marker-write step.

4. **The error message claim is now accurate.** `_block_pre` at `:531` emits "write the Rule 18 pre-implementation checkpoint block". Post-fix, this branch is reachable only when neither POST_REQ is armed NOR PRE_DONE is non-empty for the session — i.e. genuinely no pre-checkpoint exists. The instruction matches the actual missing prerequisite, addressing the original #362 secondary complaint.

5. **DL-EX-A is a high-value test.** Multi-step workflow simulators that walk `pre → post → push (block, self-arm) → push (allow, sweep convergence)` turn-by-turn are rare in this test suite. Pre-fix, the equivalent simulation would have deadlocked at step 2 (post-write blocked, misleading message). The simulator's `step3a` correctly captures that the push-gate's B1 hard-gate ALWAYS blocks the first push attempt regardless of POST_DONE state (TOCTOU-free design at `:1517-1521`). I almost flagged that as a regression — but it's the documented B1 design, also exercised by idle-push tests 6/7/40/41.

6. **Self-demonstration loop closes correctly.** This PR's session reproduced #362 against itself while authoring the post-checkpoint block (live hook was pre-fix merged-main). The workaround used was exactly the one documented in the issue: attempt push → push-gate self-arms POST_REQ + blocks → write post-checkpoint → retry push. The bug is real, observable, and recurring against the agent that's fixing it.

7. **Top-of-file lifecycle comment update is essential.** The `:38-41` documentation at HEAD-of-file is the contract future reviewers reason from. Codex caught it in round 1 as the sole FU; folding it inline (vs as a separate issue) is correct per the inline-FU heuristic — it's a 3-LOC same-surface fix and would be confusing to ship out-of-band.

### Did NOT find

- No cross-session bypass under the new predicate (`I362-4`, `DL-EX-D`).
- No zero-byte PRE_DONE bypass (`I362-3` — `[ -s ]` semantics).
- No plan-pending invariant weakening (`plan_pending_blocks_this_session` at `:1305` fires BEFORE this case-branch in the Edit/Write path, untouched).
- No wrong-root path-class regression (`_is_wrong_root_marker_write` at `:1296` fires before the basename-class predicate, untouched).
- No idempotent re-write regression (`DL-EX-E`).
- No symlink path-class regression (predicate operates on basenames AFTER canonical-root check at `:184`).

### Verdict

Recommend approval. The change is narrow (4 lines of logic + comment update), the test coverage is layered (per-axis I362-* + multi-step DL-EX-* simulator + cross-session + zero-byte + idempotent controls), and the cross-tool review (codex via second-opinion harness) returned ACCEPT-with-FU which was folded inline. Closes #362.

Hook: bot review per Rule 17 — user approves in UI.

lantisprime:
PR-level codex review complete (round 2) — **ACCEPT-with-FU**. Reply episode `20260528-060314-reply-codex-to-20260528-055830-pr-level--6cb9` walked the full 8-axis matrix (cross-hook integration, installer/runtime, cross-tool contract, symlink/path-class, cross-session, cleanup convergence, test scaffolding, negative controls). All axes EXECUTED or COVERED with locally-verifiable artifacts. Single FU (pre-existing marker-read symlink trust via `[ -s ]` follow-symlink semantics — not regression-introduced by this PR) filed as #364.

lantisprime:
PR-level codex review chain converged — **Round 5 ACCEPT** (reply episode `20260528-081006-reply-codex-to-20260528-080704-round-5-p-5352`).

Full chain on this PR (#363 closes #362, folds in #364):

| Round | Scope | Verdict | Disposition |
|---|---|---|---|
| 1 | Staged diff (#362 fix) | ACCEPT-with-FU | Stale top-of-file comment folded inline in `8193f59` |
| 2 | PR-level (8-axis matrix) | ACCEPT-with-FU | Leaf-symlink rejection on marker-read predicates folded inline in `7fa7bd8` (per the "fold in the FUs" direction) |
| 3 | R2 fix (symlink hardening) | HOLD-P1 | Writer-side `arm-if-missing` used `fs.existsSync`/`[ -e ]` (follows symlinks) → mismatched with reader-side `[ ! -L ]` precheck. Push bypass possible. Fixed in `c736a8f` (lstat-aware `isRealMarkerPresent` + `_arm_marker_via_touch_safely` self-heal pattern). |
| 4 | R3 fix (writer-side leaf) | HOLD-P1 | Same class one level up: symlinked `.checkpoints/` parent dir let writes physically land outside project root while gate reported paths inside. Repro confirmed against both helper and bash fallback paths. Fixed in `684da8f` (lstat-aware `ensurePrimaryDir` / `ensure_primary_dir` self-heal — pattern ported from `classifier-marker.mjs:validateMarkerStoreDir` precedent). |
| 5 | R4 fix (parent-dir symlink) | **ACCEPT** | No blocking findings. Trust property equivalent to `validateMarkerStoreDir` precedent; no remaining vectors at this level. |

**Tests:** 343/0 on `bash tests/test-checkpoint-gate.sh`. New regression assertions across the chain: I362-1..5, DL-EX-A/B/C/D/E (workflow simulators), I364-1..11 (symlink rejection at leaf + writer-leaf + parent-dir levels, both helper and bash fallback paths, with outside-leak negative controls).

**Push enforcement throughout:** unchanged — push-gate at `:1485` still self-arms POST_REQ unconditionally on first push attempt, `:1525` still blocks push when POST_DONE is empty. All symlink-rejection commits are READ/WRITE policy changes; the load-bearing push gate is preserved at every step.

**Disclosure:** the runbook's round-cap rule of thumb engaged at round 4 (HOLD with NEW finding); I should have surfaced to the user there rather than auto-dispatching round 5. The user re-authorized continuation ("try again, reach consensus with codex"), and the iteration did converge cleanly to ACCEPT — but the procedural miss is logged here for the record.

Awaiting your UI approval + merge.

lantiscooperdev (approved):

PR #393, opened 2026-06-17
feat(rfc-008): P3b-2 — contract-driven effective-tier layer in enforce-contract.mjs (R3/R5/R6/R8)

lantisprime:
## RFC-008 P3b-2 — contract-driven effective-tier layer in `enforce-contract.mjs`

Serves **R3, R5, R6, R8** (+ F3, CLASS-C, M8); folds in the **P4** config schema. Builds on P3b-1 (the relocated stop decision) by adding the effective-tier layer on top of the marker logic:

```
effective_tier(stop) = min( harness_cap[stop], contract.stop.tier, config.<bp>.stop )
```
over a **base-STRONG fold**.

### The anti-fail-open invariant (B1)
**default-STRONG / clamp-only-on-success.** Every tier source the boundary cannot resolve (missing contract root, missing `plugins/_index.json`, absent/unreadable `bp-001`/`events`, broken config, schema-load crash) is folded as `null` = **no clamp** → the base stays STRONG = today's unconditional refuse. A resolution failure can never weaken the gate. `enforce-contract` changes observable behavior **only** when a project's `enforce-config.json` explicitly clamps `stop` below STRONG (or sets `active:false`). Verified empirically by test-D1–D8.

### What's in it
- **`scripts/lib/effective-tier.mjs`** (NEW) — the single tier algebra, extracted from `validate-plugin-registry.mjs` (Rule 14: one algebra, two consumers). Adds the base-STRONG concrete fold (never null — F-NEW-2) and `GATE_EVENT_MAP`/`GATE_CONTRACT_KEY` (**F-NEW-1**: the stop refuse maps to the `stop` event + the `stop.tier` contract key, **not** `gates.post_checkpoint`).
- **`scripts/enforce-contract.mjs`** — `resolveContractRoot` (own depth-1 climb, mirrors P3c's two-candidate + realpath-round-trip pattern — **B3**), `resolveHarnessCap` (**M8** duplicate-binding fail-closed), `loadEnforceConfig` (**M2** full error-branch matrix → identity stay-STRONG), `CLASS-C(a)` REAL block (**N1**), `M4` downgrade notice. `decideStop` stays **pure** (gains a `stopTier` param; default STRONG keeps the em-recall parity byte-identical).
- **`patterns/enforce-config.schema.json`** (NEW, P4 folded) — `propertyNames` + schema-valued `additionalProperties` (the closed-subset validator rejects `patternProperties`).
[...]

PR #398, opened 2026-06-18
RFC-008 P4c: layer-wide active:false enforcement kill switch

lantisprime:
## RFC-008 P4c — layer-wide `active:false` enforcement kill switch (serves R5)

Closes the R5 gap: after P4a, `enforce-config.json {"active":false}` silenced only
the 4 bp-001 gates, but **not** the other ec enforcement surfaces. R5 is layer-wide —
"no enforcement plugin active → classifier silent, no hook spawns, no token cost."
P4c adds one shared, fail-closed *"is the layer active?"* consult that every
non-bp-001 ec surface honors.

### Changes
- **`enforce-contract.mjs`** — pure `layerDisposition({duplicate,active})` (M8 duplicate ⊐ `active:false`, byte-identical precedence to `gateDisposition`); new `--layer-active --marker-root <abs>` CLI (mirrors `--resolve-gate`: prints `inactive` alone or **nothing**; fail-closed on every miss/garbage/relative-root/M8/error). **F1-fold:** `--session-start` skips baseline/sweeps/advisory on `active:false` (single `resolveRepoRoot`); a schema-miss still runs them (fail-closed to *doing* the work).
- **`preflight-gate.sh`** — consult **after** the non-overridable `marker_write` + helper guards (R4 substrate-trust survives the switch) and **only** for an actual `codex-review-handoff` (no node-spawn on the ~all non-handoff tool calls). `set +e`-wrapped capture so a degraded-install ENOENT can't abort the gate into a non-blocking exit 1.
- **`second-opinion-gate.mjs`** — lazy consult at the single `emitBlock` chokepoint (fires only when a block is imminent → no spawn on allow paths, audit only on real silenced blocks). Realpath'd marker-root (no cross-resolution equality → P3b-1 isMain class can't recur); bounded spawn timeout; fail-closed on spawn failure / non-`inactive`.
- **`enforcement.md`** — §10 stale-prose fix (schema landed P3b-2 #393, not "lands in P4") + documents the `--layer-active` convention so external hooks (e.g. a user-global `compound-bash-gate.sh`) can opt in with a ~3-line read.

### Fail-closed invariant
A missing/garbage/hostile config, an M8-duplicate registry, a symlinked cwd, a
degraded install (consult ENOENT), or any resolution miss **all keep enforcement ON**.
Only an exact-string `inactive` silences a surface. `marker_write`/helper guards (R4)
survive the switch.

### Tests (CI-wired in `plan-marker-validate.yml`)
- `test-layer-active.mjs` (22) — pure algebra + closed-token CLI D1–D8 + F1 session-start fold.
- `test-layer-active-bridge.sh` (8) — preflight + second-opinion honor `inactive`; F1 ENOENT fail-closed; G4 marker-write survives; G1 spawn-fail fail-closed.
- Full regression green: preflight 107, second-opinion 23, enforce-contract 25, session-start 20, resolve-gate 27+12; `validate-bp-contract` 76 checks OK.

### Review trail
- **Plan** (2 rounds): `negative-scenario-planner` HOLD (G1–G6) → folded; `negative-scenario-reviewer` ACCEPT-WITH-FU (reproduced the `set -e` fail-open F1) → folded.
- **Code**: `negative-scenario-reviewer` ACCEPT-WITH-FU, **no required-before-merge**. All FUs (F-PERF-1/2, F-IO-1, F-OBS-1, F1-fold asymmetry) fixed inline via the lazy block-gated consult — the original F3 caching FU is obsoleted by that design.

### Post-merge follow-ups (separate)
- Deploy to global hooks: `install.mjs --tool claude-code --install-hooks --install-hooks-force`.
- RFC-008 `README.md` phase-index + workplan update.

### DIFF
--- a/.github/workflows/plan-marker-validate.yml
+++ b/.github/workflows/plan-marker-validate.yml
@@ -99,6 +99,12 @@ jobs:
       - name: Run --resolve-gate bash bridge contract (RFC-008 P4a, F5 — exact-token match, ||-envelope fail-closed, plan-gate honors silence/clamp)
         run: bash tests/test-resolve-gate-bridge.sh
 
+      - name: Run --layer-active per-project kill switch (RFC-008 P4c, R5 — layerDisposition + closed-token CLI + F1 session-start fold)
+        run: node tests/test-layer-active.mjs
+
+      - name: Run --layer-active bash+node bridge (RFC-008 P4c, R5 — preflight + second-opinion honor inactive; F1 ENOENT fail-closed; G4 marker-write survives)
+        run: bash tests/test-layer-active-bridge.sh
+
       - name: Run enforce-contract --session-start side-effects relocation (RFC-008 P3d)
         run: node tests/test-enforce-contract-session-start.mjs
 
--- a/plugins/claude-code/hooks/preflight-gate.sh
+++ b/plugins/claude-code/hooks/preflight-gate.sh
@@ -296,6 +296,32 @@ if [ "$CLAIM_CLASS" != "codex-review-handoff" ]; then
   exit 0
 fi
 
+# ---------------------------------------------------------------------------
+# RFC-008 P4c (R5): layer-wide enforcement kill switch.
+# Placed AFTER the claim-class early-exit (F-PERF-1) so it spawns the consult ONLY
+# for an actual codex-review-handoff — the one enforcement this gate applies and
+# the only thing the switch silences — never for the ~all non-handoff Write/Edit/
+# Bash/Agent calls. It also sits AFTER the non-overridable marker-write (:136-195)
+# + helper-invocation (:216-283) guards, which protect substrate trust (R4:
+# marker_write NON-overridable) and MUST survive active:false (planner G4).
+#
+# FAIL-CLOSED capture (reviewer F1): the script runs under `set -e` (line 2), so a
+# naive `OUT="$(node …)"` that ENOENTs (degraded install) would ABORT the gate →
+# exit 1 → PreToolUse non-blocking → tool proceeds UN-gated (the exact inverse of
+# the kill switch). Capture inside an explicit `set +e` block (mirror the
+# canonicalize pattern at :142-151) and skip enforcement ONLY on the exact string
+# `inactive`. Every other outcome — non-zero exit, ENOENT, empty, stderr-only, any
+# non-`inactive` string — falls through to enforcement (2>/dev/null discards the
+# node stderr so it can never leak into the captured token).
+# ---------------------------------------------------------------------------
+LAYER_ACTIVE_CONSULT="$HOME/.episodic-memory/scripts/enforce-contract.mjs"
+set +e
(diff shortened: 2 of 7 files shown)


lantiscooperdev (commented):
## PR-level review — RFC-008 P4c (full `main...HEAD`, integration layer)

Third review layer (after per-slice planner + per-artifact code review). Focus: cross-cutting / integration concerns the narrower passes don't surface. Verified against the integrated diff + a live fail-closed sweep on disk.

### Integration checks — PASS
- **Single source of truth.** Both new surfaces route through the *same* global `enforce-contract --layer-active` with an exact-string `inactive` match (`preflight-gate.sh` via `set +e` bash capture; `second-opinion-gate.mjs` via `spawnSync` at the `emitBlock` chokepoint). No divergent decision logic.
- **CLI dispatch ordering.** The `--layer-active` branch sits after `--resolve-gate` and before the `--gate stop` required-check. A `--layer-active` invocation (no `--gate`) is handled before the gate check; `--gate stop` / `--session-start` / `--resolve-gate` are unaffected. No mode interference.
- **M8 ⊐ active precedence** is byte-identical to `gateDisposition` (`layerDisposition` unit A3/A4 + CLI B8). A corrupt/duplicate registry is non-silenceable.
- **F1-fold asymmetry** (session-start checks raw `cfg.active`, no M8) is deliberate and fail-closed — a skipped baseline only removes an allow-arming carve-out the still-blocking stop gate ignores. Documented inline.
- **Lazy consult** (second-opinion) + **claim-class-gated consult** (preflight) mean the node-spawn fires only when a block is imminent — no per-tool-call cost on the active path, and the audit line lands only on a real silenced block.

### Fail-closed invariant — CLOSED on every surface
Missing/garbage/hostile config, M8-duplicate registry, relative/missing marker-root, consult ENOENT, spawn failure/timeout, symlinked cwd, and the **old deployed global** (no `--layer-active` branch → error JSON → non-`inactive`) all keep enforcement ON. Only an exact `inactive` silences. R4 `marker_write`/helper guards survive the switch (bridge G4).

### Tests
22 (CLI/algebra/F1) + 8 (bridge) new, CI-wired in `plan-marker-validate.yml`. Full regression green: preflight 107, second-opinion 23, enforce-contract 25, session-start 20, resolve-gate 27+12; `validate-bp-contract` 76 OK (§10 prose edit is above the M7f `CONFIG:BEGIN` block, so the byte-diff is untouched).

### One integration dependency (operational, not a code defect)
The `--layer-active` CLI + the two hook edits only take effect once deployed to the global hooks/scripts the running hooks actually invoke:

```
node install.mjs --tool claude-code --install-hooks --install-hooks-force
```

Until then the surfaces fail-closed (enforce) against the stale global — verified, safe — but the kill switch is inert. This is the standard "merge ≠ deployed" closing step, tracked in the PR description's post-merge follow-ups.

### Verdict
**LGTM** — no required-before-merge code changes. Fail-closed spine is sound and proven on disk; all prior-round findings (G1–G6, F-PERF-1/2, F-IO-1, F-OBS-1) are resolved. Posting as a comment per Rule 17 — user approves in the UI.

lantiscooperdev (approved):

PR #400, opened 2026-06-19
test(rfc-008): P4d S1 — mock-project activation-scoping E2E harness

lantisprime:
## RFC-008 P4d — Slice S1: mock-project activation-scoping E2E harness

First slice of **P4d** (decouple the enforcement layer from the memory substrate). Establishes the mock-project E2E fixture lib reused by S2–S8 and the first test, **green on current code** — the foundation later slices assert against.

### What this proves (RQ3, Principle 12)
A **core install ships the substrate + activation-gated BP-1 hygiene with ZERO enforcement gates**, and every em-* round-trip works with no enforcement hook present. Empirically verified by running the REAL `install.mjs` / REAL deployed scripts / REAL hooks against an isolated HOME + mock git project (no mental tracing).

### Contracts (13/13 green)
- **A0–A4** install hook-scoping baseline: core install writes **no global `settings.json`**; project `SessionStart` holds **only the two BP-1 hooks**; substrate scripts deployed; BP-1 hook exits 0 silently when not activated; **A4** is a Rule-14 drift guard asserting `ENFORCEMENT_HOOK_MARKERS` covers every `HOOK_SPECS` hook + `SESSION_END_SCRIPT`.
- **C1–C7** substrate round-trips hook-free: em-store / search / revise / rebuild / list / recall (local + global), supersession honored, recall side-effect-free.

### Files
- `tests/lib/activation-scoping-harness.mjs` — `mkMock`, `runInstall`, `runScript`, `runHook`, `readSettings` + enforcement-gate detection (one `scrubEnv` for HOME isolation).
- `tests/test-activation-scoping-e2e.mjs` — the S1 contracts.
- `.github/workflows/plan-marker-validate.yml` — one CI step.

### Review
Second-opinion (claude-subagent) code review: **ACCEPT, no blocking findings**. Two non-blocking foundation gaps (F1 marker completeness, F2 env scrub) folded **inline**; F3 (enforce-config writer + block-vs-allow gate-runner) tracked into the **S4** slice plan.

### Scope notes
Test-only — **no production code touched**. Does **not** yet exercise `--install-enforcement` (that is S2). Principle 12 + the RFC P4d doc land in later slices.

### DIFF
--- a/.github/workflows/plan-marker-validate.yml
+++ b/.github/workflows/plan-marker-validate.yml
@@ -123,6 +123,9 @@ jobs:
       - name: Run install-runtime contract-set deploy coupling (RFC-008 P3b-2, §10)
         run: node tests/test-install-contract-deploy.mjs
 
+      - name: Run activation-scoping mock-project E2E (RFC-008 P4d S1 — substrate hook-independence A0-A3 + em-* round-trips C1-C7)
+        run: node tests/test-activation-scoping-e2e.mjs
+
       - name: Run preflight-gate regression tests
         run: bash tests/test-preflight-gate.sh
 
--- a/tests/lib/activation-scoping-harness.mjs
+++ b/tests/lib/activation-scoping-harness.mjs
@@ -73,6 +73,9 @@ process.on('exit', () => {
     try { fs.rmSync(d, { recursive: true, force: true }) } catch {}
   }
 })
+// 'exit' doesn't fire on a forced signal (local SIGINT, CI timeout SIGTERM);
+// re-exit through process.exit so the cleanup above runs (S1 PR-review F3).
+for (const sig of ['SIGINT', 'SIGTERM']) process.on(sig, () => process.exit(130))
 
 /**
  * Create an isolated fixture: a fresh HOME, a mock git project, and a separate
--- a/tests/lib/activation-scoping-harness.mjs
+++ b/tests/lib/activation-scoping-harness.mjs
@@ -28,20 +28,44 @@ export const REPO_ROOT = path.resolve(
   path.dirname(new URL(import.meta.url).pathname), '..', '..'
 )
 
-// Command-string fragments that identify an RFC-008 ENFORCEMENT gate (the
-// --install-hooks / --install-second-opinion bundle). BP-1 SessionStart hooks
-// (bp1-approval-check / bp1-sweep-on-session) are SUBSTRATE hygiene
-// (activation-gated, RFC-004) and are deliberately NOT in this set — their
+// Command-string fragments that identify an RFC-008 ENFORCEMENT hook.
+//
+// Rule (one unambiguous definition, S1 review F1): a marker names EVERY hook
+// that `--install-hooks` / `--install-second-opinion` registers — i.e. the
+// install-manifest `HOOK_SPECS` bundle (all of which are enforcement; no
+// substrate hook lives in HOOK_SPECS) + the SessionEnd script + the
+// second-opinion gate (registered by --install-second-opinion, outside
+// HOOK_SPECS). The BP-1 SessionStart hooks (bp1-approval-check /
+// bp1-sweep-on-session) are wired by a SEPARATE install path, are SUBSTRATE
+// hygiene (activation-gated, RFC-004), and are deliberately absent here — their
 // project-scoped presence is expected and correct.
+//
+// test-activation-scoping-e2e.mjs (A4) asserts this set covers every HOOK_SPECS
+// file + SESSION_END_SCRIPT, so it cannot silently drift behind install (Rule 14).
 export const ENFORCEMENT_HOOK_MARKERS = [
   'plan-gate',
   'checkpoint-gate',
   'preflight-gate',
   'stop-gate',
-  'second-opinion-gate',
   'em-recall-sessionstart',
+  'preflight-prompt-helper',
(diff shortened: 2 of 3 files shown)


lantiscooperdev (commented):
## Review — RFC-008 P4d Slice S1 (activation-scoping E2E harness)

Reviewed the diff against the locked P4d requirement (**enforcement per-project, never global; substrate stays global + hook-free**, Principle 12 / RQ3/RQ7) and for harness soundness as the foundation S2–S8 build on.

### Strengths
- **Empirically grounded, not mental-traced.** Every primitive was validated by running the real `install.mjs` / deployed scripts / hooks under an isolated HOME + mock git project. The harness encodes two facts a code-read would miss: `--project` is a namespace label (local scope = cwd repo root, so local ops pin `cwd: M.project`), and macOS `/tmp`→`/private/tmp` (base is realpath'd).
- **Load-bearing assertions.** A1g (`global settings === null`) and A1p (exact-shape `deepStrictEqual` on the hook tree) catch any enforcement leak regardless of the marker list; C1–C7 fail loudly on null/garbled JSON. No vacuous passes.
- **A4 drift guard (Rule 14).** Asserts `ENFORCEMENT_HOOK_MARKERS` covers every `HOOK_SPECS` hook + `SESSION_END_SCRIPT`, so the marker set can't silently fall behind `install`. It is load-bearing — it would have failed before the F1 additions below.

### Second-opinion findings (claude-subagent), disposition
- **F1 — marker completeness (ACCEPT, folded inline):** `ENFORCEMENT_HOOK_MARKERS` originally omitted `session-handoff-prompt` + `preflight-prompt-helper`, both registered into the global bundle by `--install-hooks`. S1 itself was unaffected (exact-shape asserts), but `hasEnforcementHook` — the helper **S2** will use to prove "no enforcement leaks to global" — would have passed green-but-wrong if a regression leaked only those two. Fixed: both markers added, classification rule rewritten to one unambiguous definition, A4 guard added.
- **F2 — env scrub (ACCEPT, folded inline):** `runHook`/`runScript` scrubbed only `CLAUDE_CONFIG_DIR`; the second-opinion gate also reads `SO_INSTALL_SNAPSHOT_PATH`/`SO_RUNBOOK_PATH`/`SO_QUICKREF_PATH`. Centralized all three runners onto one `scrubEnv()` so a stale `SO_*` can't defeat HOME isolation when S2+ exercises the gate.
- **F3 — forward note (deferred to S4):** the lib still needs an `enforce-config.json` writer + a block-vs-allow gate-runner; tracked in the S4 slice plan so S4 extends the lib instead of forking it.

### Locked-requirement compliance
Confirmed: no global-enforcement assumption anywhere in the harness or assertions. A1g/A1p prove enforcement is never global and the project owns its (BP-1) hooks; C1–C7 prove the substrate round-trips hook-free including the global store under isolated HOME. RQ3/RQ7 + Principle 12 honored.

**Verdict: ACCEPT** — no blocking findings; both non-blocking findings folded inline. Test-only, no production code touched. Approving is the user's in the UI (bot posts findings only).

lantiscooperdev (commented):
## PR-level review — RFC-008 P4d Slice S1

A second review **layer** (PR-level), distinct from the earlier per-artifact/diff pass — per the "review layers compound" lesson, hunting cross-file / CI / isolation / slice-boundary classes a line-level review misses. Verified live (real machine), not mental-traced.

**Independently confirmed:**
- **Live green + no pollution:** 13/13 on a real `$HOME`; real `~/.episodic-memory` has none of the test's `s1rt`/`s1global` tags; temp dirs all cleaned.
- **CI is a real merge gate:** the new step is in the `validate` job (~20 sibling contract tests), `validate` passed on the PR, no `paths:` filter, `process.exit(1)` failure propagation confirmed.
- **No silent scope:** `git diff --stat main...HEAD` touches only `tests/**` + 3 lines of workflow yaml — matches the "no production code touched" claim exactly.
- **Cross-file import sound:** `install-manifest.mjs` is a pure data/builder module (no cycle, no load-order issue); a legit new `HOOK_SPECS` hook makes A4 fail loudly and correctly.
- **Slice boundary clean:** S1 stops at registration/round-trip primitives; doesn't reach into S2's `--install-enforcement`; F3 deferral to S4 is recorded in the plan.
- **Locked-requirement compliant:** nothing in the PR (code/CI/commit framing) introduces a global-enforcement assumption.

**Findings (non-blocking), disposition:**
- **F1** — A4 drift guard covers `HOOK_SPECS` + SessionEnd but not the globally-registered `second-opinion-gate` (registered by `--install-second-opinion`, outside `HOOK_SPECS`). → recorded as S2 forward-debt (S2 extends the guard to that registration surface).
- **F2** — S2 must not mistake that pre-existing global `second-opinion-gate` (cross-tool review tooling) for a P4d enforcement leak. → recorded in the S2 plan-of-record.
- **F3** — `process.on('exit')` cleanup didn't fire on SIGKILL/SIGTERM (immaterial in CI). → **folded inline** (`e55abb1`): SIGINT/SIGTERM re-exit through `process.exit` so cleanup runs.

**Verdict: ACCEPT** (consensus) — no blocking findings; F3 folded inline, F1/F2 tracked into S2. Both review layers (per-artifact + PR-level) now clear. Approving is the user's in the UI.

lantiscooperdev (approved):

PR #401, opened 2026-06-19
feat(rfc-008): P4d S2 — per-project enforcement install (--install-enforcement) + substrate allowlist

lantisprime:
## RFC-008 P4d S2 — per-project enforcement install (`--install-enforcement`)

Decouples the **enforcement layer** from the **memory substrate** (RFC-008 P4d core; PRINCIPLES.md §12). This PR is S2 of the P4d slice plan: it relocates ALL enforcement install per-project and makes the global `~/.episodic-memory/` substrate enforcement-clean.

### The three governing P12 invariants

| Invariant | Status after this PR |
|---|---|
| **1. Enforcement separate from substrate** | ✅ engine + classifier + markers + bp1 + contract config relocated per-project; global is substrate-only |
| **2. Per-project on/off switch provisioned** | ⏳ **S4** — install does not yet write `enforce-config.json`; P4c already reads it |
| **3. All enforcement files/scripts/configs at project level** | ✅ files + scripts + configs all per-project |

### What changed

**Relocation (P12 function test — a thing that exists only to run/serve enforcement is enforcement, however packaged → per-project):**
- The enforcement **engine** (`enforce-contract.mjs`), the **classifier** suite, the **marker** writers, the **bp1** orchestration + their transitive `scripts/lib/*.mjs` closure + the **contract config** (taxonomy / bp-001 / events / enforce-config.schema) install under `<project>/.claude/hooks/` (+ `/patterns`, `/plugins`).
- Gate hooks (`*-gate.sh`, `em-recall-sessionstart.sh`, classifier `.sh` libs) resolve via `$HOOK_DIR` (their own dir) with the global path as an **inert legacy fallback**. All 5 gates **fail-closed** when the engine is absent at both locations; R4 non-overridable marker write preserved.
- **BP-1** (RFC-004 auto-pilot) is a behavior pattern: it ships per-project on **core** install (not gated on `--install-enforcement`), self-contained (hooks + bp1 scripts + closure).

**Substrate allowlist (final commit — closes a leak the relocation left open):**
- The global scripts scan was a **denylist** ("any `.mjs` not enforcement → global"), silently leaking 12 repo-dev/CI validators (`validate-*`, `scaffold-bp`, `test-plugin`, `check-automode-defaults`) into the substrate — including enforcement tooling like `validate-bp-contract` / `validate-plan-marker-sites`.
- Flipped to an **allowlist**: global = substrate ONLY (`em-*` + the `second-opinion` capability). Repo-dev tools ship nowhere (CI runs them repo-relative). Global drops **33 → 21** substrate-only scripts.

### Guardrail (Rule 13/14)
`tests/test-p12-global-clean.mjs` — manifest-derived, asserts across all 6 non-`--install-enforcement` variants that global scope has ZERO enforcement files/registrations/runtime AND nothing but substrate scripts. Bites pre-S2; a newly added enforcement/dev script fails CI rather than leaking.

### Review
Second-opinion (`claude-subagent`, grounded against the tree): **ACCEPT-WITH-FOLLOWUPS, zero blocking.** F1 (stale comments), F2 (`--install-hooks` no-op notice), F3 (closure regex missed bare side-effect imports + regression test) **fixed inline**. F4 (3 `em-*` validators kept global by prefix) **deferred** — verified benign (no enforcement-lib pinned via their closures).

### Tests
Full suite **40/0**, including the new `test-lib-closure.mjs` (import-form coverage).

### Scope notes
- **S2 grew beyond its original charter**: it absorbed the full engine/config relocation + the substrate allowlist (originally these were to be a later slice). Invariant 2 (provision `enforce-config.json`) remains **S4**.
- Transitional: `--install-enforcement` is the per-project enforcement entry; `--install-hooks` alone now emits a notice and deploys no enforcement (S3 resolves the public-flag story).

### DIFF
--- a/.claude/hooks/bp1-approval-check.sh
+++ b/.claude/hooks/bp1-approval-check.sh
@@ -79,6 +79,9 @@ fi
 # under <project>/.claude/hooks/, NOT in the global substrate. BASH_SOURCE is
 # absolute (Claude Code registers hooks by absolute path), so this is cd-safe.
 EM_SCRIPTS_DIR="$(cd -P "$(dirname "${BASH_SOURCE[0]}")" 2>/dev/null && pwd)"
+# Co-located install (P4d) first; legacy global fallback if the bp1 scripts are
+# not beside this hook (fresh installs co-locate; the fallback is inert there).
+[ -f "$EM_SCRIPTS_DIR/bp1-flag-check.mjs" ] || EM_SCRIPTS_DIR="$HOME/.episodic-memory/scripts"
 FLAG_CHECK="$EM_SCRIPTS_DIR/bp1-flag-check.mjs"
 VALIDATE="$EM_SCRIPTS_DIR/bp1-marker-validate.mjs"
 EMIT_INVALID="$EM_SCRIPTS_DIR/bp1-emit-marker-invalid-evidence.mjs"
--- a/.claude/hooks/bp1-approval-check.sh
+++ b/.claude/hooks/bp1-approval-check.sh
@@ -55,9 +55,9 @@ set -e
 #     WORKTREE root (not the main checkout). This hook follows the same
 #     resolution — see the $TOPLEVEL computation below.
 #
-# Script resolution (codex code-review A1, 2026-05-07):
-#   bp1 scripts are installed GLOBALLY at $HOME/.episodic-memory/scripts/.
-#   See bp1-sweep-on-session.sh:23-30 for rationale.
+# Script resolution (RFC-008 P4d / Principle 12 — relocated per-project 2026-06-19):
+#   bp1 scripts install CO-LOCATED with this hook under <project>/.claude/hooks/,
+#   resolved via $HOOK_DIR. NOT global. See bp1-sweep-on-session.sh:23-29.
 
 INPUT="$(cat)"
 # Guard stdin .cwd parse: malformed SessionStart JSON would otherwise propagate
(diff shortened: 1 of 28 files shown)


lantiscooperdev (commented):
## Agent review — RFC-008 P4d S2 (per-project enforcement install + substrate allowlist)

Reviewed against the actual tree (no mental tracing) + full suite execution. Posting findings here per Rule 17 so they're visible in the Reviews tab; **user approves in UI**.

### Verified clean
- **Fail-closed:** all 5 gates (stop / plan / checkpoint / preflight / em-recall-sessionstart) deny when the engine is missing at BOTH co-located (`$HOOK_DIR`) and the inert global fallback. R4 non-overridable marker write survives (checkpoint-gate resolves the helper co-located→global before the disposition consult).
- **`HOOK_DIR` defined before first use** in every gate.
- **Partition correctness:** global = 21 substrate-only scripts (`em-*` + `second-opinion`); 0 enforcement, 0 repo-dev validators leaked. `relocatedOnlyLibs = enforcementBundleLibs − globalScriptLibs` holds by construction. No substrate `em-*` script spawns an excluded sibling (git/gh/self only).
- **Guardrail bites:** `test-p12-global-clean` is manifest-derived across all 6 non-`--install-enforcement` variants; a newly added enforcement/dev script fails CI rather than leaking.
- **Full suite 40/0**, incl. the new `test-lib-closure.mjs`.

### Findings (from the second-opinion pass, all dispositioned)
- **F1** stale comments describing the deleted global taxonomy/contract deploy — **fixed** (`install.mjs`).
- **F2** `--install-hooks` w/o `--install-enforcement` was a silent enforcement no-op — **fixed** (one-line notice, mock-verified).
- **F3** `computeLibClosure` missed bare side-effect `import './x'` (latent; the per-project bundle's sole completeness guarantee) — **fixed** + regression test.
- **F4** 3 `em-*` validators kept global by prefix — **deferred**, verified benign (no enforcement-lib pinned via their closures).

### Scope flag (not a blocker)
S2 grew to absorb the full engine/config relocation + the substrate allowlist. **Invariant 2** (install provisions `<project>/.episodic-memory/enforce-config.json`) is **not** in this PR — it's S4. So the per-project on/off switch is mechanism-present (P4c reads it) but not yet provisioned. Calling that out explicitly so it isn't mistaken for complete.

**Verdict: ACCEPT-WITH-FOLLOWUPS** (no blocking issues). Recommend merge after UI approval.

lantiscooperdev (approved):

PR #404, opened 2026-06-19
chore(deploy): add tools/deploy-audit.mjs — unfiltered post-merge deploy verification

lantisprime:
## What

Adds `tools/deploy-audit.mjs` — an authoritative, **unfiltered** post-merge deploy check, to replace leaning on `migration-cutover.mjs` for "is global deployed?"

## Why

`migration-cutover.mjs` verifies the global install against the install **manifest**, which only lists the `.mjs` substrate closure (em-* + second-opinion + lib closure). It is **structurally blind** to every other file class. When install logic narrows over time — the S2 denylist→allowlist flip, or dropping a `.sh`/`.tmpl` copy — files deployed by the *old* logic persist in `~/.episodic-memory/` as stale orphans the cutover can't see.

Concretely (RFC-008 P4d S4 follow-up): a "cutover 37/37" pass got rounded up to "global is clean", but an unfiltered audit found **4 stale `.sh`/`.tmpl` orphans** from a pre-allowlist install (`em-skill-wrapper.sh.tmpl`, `render-input-validation.sh`, two `-test.sh`). They've since been pruned; this tool makes that check repeatable.

## What it does

Clean install into a throwaway HOME (via the canonical `activation-scoping-harness`, so it inherits the env-scrub that keeps the audit honest), then byte-diff **every** file in the four global-deployed dirs:

- `MISSING` — in a clean install, absent from real global → an update never deployed
- `DIFFER` — present in both, bytes differ → real global stale
- `EXTRA` — in real global, not in a clean install → orphan / stale leftover

**Report-only by design** — it never deletes. An "orphan" can be a feature file whose consumers must be traced first (`render-input-validation.sh` *looked* prunable but is the launchd-routines install-time validator). Cosmetic DIFFERs (the second-opinion snapshot's `install_timestamp`; content hash is invariant) are demoted to info.

Inherently a **local/ops tool** (audits the developer's real `~/.episodic-memory` + `~/.claude/hooks`) — **not** CI-wired; CI has no real global to audit. Lives in `tools/` (repo-dev scope) so the installer never deploys it.

```
$ node tools/deploy-audit.mjs
  (info)   .claude/hooks/second-opinion-providers.json  cosmetic DIFFER (install_timestamp only; content current)

MISSING=0  DIFFER=0  EXTRA=0  (+1 cosmetic, ignored)
CLEAN — real global == authoritative clean install.
```

Modes: default, `--json`, `--help`. Exit 0 clean / 1 drift / 2 harness failure.

### DIFF
--- a/tools/deploy-audit.mjs
+++ b/tools/deploy-audit.mjs
@@ -78,6 +78,17 @@ function walk(root) {
   return out.sort()
 }
 const byteEqual = (a, b) => { try { return fs.readFileSync(a).equals(fs.readFileSync(b)) } catch { return false } }
+// A COSMETIC_DIFFER is cosmetic ONLY when the SOLE difference is the non-semantic
+// `install_timestamp` — a real `source_hash` (provider-registry) change must still
+// surface as DIFFER, never be swallowed (review F1). Unparseable → treat as real DIFFER.
+function snapshotDiffIsOnlyTimestamp(realPath, mockPath) {
+  try {
+    const a = JSON.parse(fs.readFileSync(realPath, 'utf8'))
+    const b = JSON.parse(fs.readFileSync(mockPath, 'utf8'))
+    delete a.install_timestamp; delete b.install_timestamp
+    return JSON.stringify(a) === JSON.stringify(b)
+  } catch { return false }
+}
 
 const REAL = os.homedir()
 const M = mkMock('deploy-audit')
@@ -100,8 +111,10 @@ for (const d of DIRS) {
   for (const f of mockFiles) {
     const key = `${d}/${f}`
     if (!realSet.has(f)) { findings.missing.push(key); continue }
-    if (!byteEqual(path.join(realDir, f), path.join(mockDir, f))) {
-      (COSMETIC_DIFFER.has(key) ? findings.cosmetic : findings.differ).push(key)
+    const rp = path.join(realDir, f), mp = path.join(mockDir, f)
+    if (!byteEqual(rp, mp)) {
+      if (COSMETIC_DIFFER.has(key) && snapshotDiffIsOnlyTimestamp(rp, mp)) findings.cosmetic.push(key)
+      else findings.differ.push(key)
     }
   }
   for (const f of walk(realDir)) {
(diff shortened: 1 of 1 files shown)


lantiscooperdev (commented):
**Bot review (lantiscooperdev) — ACCEPT-WITH-FOLLOWUPS, no blocking** (user approves via UI).

Reviewed `tools/deploy-audit.mjs` against its purpose + the harness it reuses.

**Correctness**
- Reuses `activation-scoping-harness` `mkMock`/`runInstall` — the right call: it inherits `scrubEnv` (CLAUDE_CONFIG_DIR + SO_* stripping), without which the audit would compare against a HOME-polluted install and silently lie. A naive inline runner would have missed that.
- The three-way classification is sound: MISSING (clean-install file absent in real → undeployed), DIFFER (bytes differ → stale), EXTRA (real file absent from clean install → orphan). Symmetric set-diff over a recursive walk of the four global dirs.
- `COSMETIC_DIFFER` correctly demotes the second-opinion snapshot (per-install `install_timestamp`; `source_hash` invariant) to info so it doesn't mask as drift — and doesn't suppress a *real* content change, since only that one path is exempted and a content delta would still alter bytes the audit reports… actually it would be swallowed too. See follow-up.

**Report-only** is the right safety posture — the `render-input-validation.sh` case (orphan that was actually the launchd install-time validator) is exactly why auto-prune would be dangerous. Good that it's documented inline.

**Scope** — lives in `tools/` (repo-dev), not in the substrate allowlist, so the installer never deploys it; not CI-wired (correct — CI has no real global). Clean.

**Follow-up (non-blocking):**
- F1: `COSMETIC_DIFFER` swallows *any* byte difference in `second-opinion-providers.json`, including a genuine `source_hash` change. Low risk (a real provider-registry change is caught by the dedicated second-opinion install E2E), but tightening to compare-after-stripping-`install_timestamp` would make it exact. Defer.

Verdict: ACCEPT-WITH-FOLLOWUPS. Recommend approve + merge.

lantiscooperdev (approved):

PR #397, opened 2026-06-18
feat(rfc-008): P4a — per-project enforce-config wired into the 3 pre_tool_use gates

lantisprime:
## RFC-008 P4a — per-project `enforce-config.json` wired into the 3 pre_tool_use gates

Extends per-project `enforce-config.json` (stop-gate-only since P3b-2 #393) to the three
`pre_tool_use` classification gates — `plan_approval`, `pre_checkpoint`, `post_checkpoint`.
Serves **R3** (`effective_tier = min(harness, contract, project_config)`, clamp DOWN only),
**R5** (`active:false` → silence), **R6** (M8 one-harness-one-plugin parity).

### Design — consult-before-block
- `enforce-contract.mjs`: new `--resolve-gate <gate> --marker-root <abs>` mode + pure
  `gateDisposition()`. Emits a CLOSED token vocab to stdout — `silence` | `clamp-off` — and
  nothing else; the bash gate BLOCKS by default and skips its block ONLY on an exact-equality
  safe token. `--marker-root` is passed explicitly by the hook (no re-resolve → the P3b-1
  `/var` isMain fail-open class cannot recur).
- `effective-tier.mjs`: `LIVE_GATES` → all four gates.
- `plan-gate.sh` / `checkpoint-gate.sh`: a single early consult per gate (spawns only when a
  block is imminent). `LABEL → gate` selection (push → `post_checkpoint`, else `pre_checkpoint`);
  `post_checkpoint` silence routes through the existing marker-cleanup sweep so
  `.checkpoint-required` is never stranded → **no stop-gate deadlock**; integrity +
  marker_write blocks stay **non-silenceable**.
- `validate-plugin-registry.mjs` + `runbooks/enforcement.md`: `enforce_config_keys` derived
  from `GATE_CONTRACT_KEY` (Rule-14 single source, M7f byte-match).

### Fail-closed (B1)
`node` absent / non-zero exit / empty / garbage / substring-not-equal → the gate keeps
blocking. M8 duplicate-plugin and integrity blocks are non-silenceable even under `active:false`.

### Reviews
- **Plan:** 2-round second-opinion (claude-subagent) — HOLD→HOLD→converged (F1–F12).
- **Code:** `negative-scenario-reviewer` → **ACCEPT-with-FU**, 0 blockers / 0 major. Every
  fail-open axis verified fail-**closed** with real shell traces. All 3 findings resolved in-PR:
[...]

PR #406, opened 2026-06-20
fix(rfc-008): P4d — prune superseded relative-path hook duplicate on re-install

lantisprime:
## Summary

RFC-008 P4d fix: a prior installer registered enforcement hooks by **project-relative** path (`.claude/hooks/X.sh`); the current installer registers the canonical **absolute** path. `addHookEntry` keys idempotence on the exact command *string*, so the two spellings never unify — re-installing left **both** entries and every gate hook fired **twice**.

## Changes

- **`install.mjs`** — add `hookCommandTargetPath()` (resolve a hook command's script to an absolute path, stripping the `bash`/`node` runner + quotes) and `removeHookEntryByCommand()`. On re-install, for each stale canonical-basename entry:
  - resolves to the **same canonical file** → **auto-remove** the superseded spelling (canonical absolute entry kept);
  - shares only the **basename** but resolves elsewhere → **warn only, never remove** (preserves the operator's own hook — original safety intent).

  Previously the code *only ever warned*, so duplicates were never pruned. Also refreshes `--help` for the P4d per-project model (`--install-enforcement`).
- **`tests/test-install-dedup.mjs`** (new) — real `install.mjs` + mock project; asserts the superseded relative spelling is auto-removed and a different-path custom entry is preserved.
- **`tests/test-plan-gate.sh`** — thread a valid `session_id` into the mock JSON (P4a/P4d #397/#401 made plan-gate fail-closed without one, which would have blocked 8 read-only assertions) + add tests 11/12 for the missing-session_id fail-closed boundary.
- **`.github/workflows/plan-marker-validate.yml`** — wire `test-plan-gate.sh` (was never in CI) + `test-install-dedup.mjs` into the validate job.

## Test plan

- `node tests/test-install-dedup.mjs` → **8/0** (real install, mock project, isolated HOME)
- `bash tests/test-plan-gate.sh` → **36/0** (hardened, session_id-bearing)
- Both now run in CI via `plan-marker-validate.yml`.

## Notes

Independent of the in-flight ESC (enforcement-scope-correction) work — touches `install.mjs` + the workflow, which ESC-S3 also touches, so whichever lands first, the other rebases.

### DIFF
--- a/.github/workflows/plan-marker-validate.yml
+++ b/.github/workflows/plan-marker-validate.yml
@@ -84,8 +84,14 @@ jobs:
       - name: Run checkpoint-gate regression (incl. F1 Bash re-gate + .review-store carve-out, PR-B)
         run: bash tests/test-checkpoint-gate.sh
 
+      - name: Run plan-gate regression (incl. P4a/P4d session_id fail-closed guard)
+        run: bash tests/test-plan-gate.sh
+
       - name: Run install-hooks regression (incl. PR-B orphan sweep + manifest count)
         run: bash tests/test-install-hooks.sh
+
+      - name: Run install dedup regression (RFC-008 P4d double-registration prune)
+        run: node tests/test-install-dedup.mjs
 
       - name: Run stop-gate regression
         run: bash tests/test-stop-gate.sh
--- a/install.mjs
+++ b/install.mjs
@@ -114,7 +114,7 @@ if (bootstrapLastPrompt) {
 }
 
 if (!tool) {
-  console.log(`Usage: node install.mjs --tool <claude-code|cursor|codex|opencode|pi-agent|windsurf|all> [--project <path>] [--install-hooks] [--install-hooks-force]
+  console.log(`Usage: node install.mjs --tool <claude-code|cursor|codex|opencode|pi-agent|windsurf|all> [--project <path>] [--install-hooks] [--install-enforcement] [--install-hooks-force] [--install-second-opinion] [--bootstrap-last-prompt]
 
 Tools:
   claude-code  Install SKILL.md + plugin structure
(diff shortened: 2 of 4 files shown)


lantiscooperdev (commented):
## Bot review — automated pre-merge pass (not an approval)

Reviewed the dedup logic adversarially + ran the regression locally (`test-install-dedup.mjs` 8/0, `test-plan-gate.sh` 36/0).

### Correctness — looks sound
- `hookCommandTargetPath()` correctly handles both spellings: `node /abs/X.sh --flag` (runner stripped → absolute token) and a bare project-relative `.claude/hooks/X.sh` (resolved against `projectDir`). The same-file equality (`=== canonicalResolved`) is the right discriminator for remove-vs-warn.
- The **warn-only-when-resolves-elsewhere** branch preserves an operator's own same-basename hook — the original safety intent is intact, and the new behavior only prunes a *provably superseded* spelling of the canonical file.
- Idempotence: after the prune, exactly one canonical absolute entry remains (asserted).

### Findings

**F1 (NIT, low) — `removeHookEntryByCommand` removes the whole entry block.** If a stale relative command ever shared a single hook-entry block with *other* hooks (e.g. an operator hand-merged matchers), filtering the entry would drop those siblings too. Installer-written entries are always one-hook-per-entry, so this can't arise from our own output — but a one-line guard (only remove when the entry's `hooks` array is exactly the single matched command) would make it robust against hand-edited settings. Non-blocking.

**F2 (NIT) — env-prefixed commands.** `hookCommandTargetPath` strips only a leading `bash|node|sh`; a command like `VAR=1 bash X.sh` wouldn't resolve. We never emit such commands, so this is theoretical. Non-blocking.

### Tests / CI
- New `test-install-dedup.mjs` drives the real installer against a mock project (no mental-trace) and covers both the auto-remove and the preserve-custom paths. 
- Good catch wiring `test-plan-gate.sh` into CI — it was passing locally but never gated.

### Verdict
No blockers. F1/F2 are optional hardening. Recommend merge after human approval.

lantiscooperdev (approved):

Issue #415, opened 2026-06-21
RFC-008 P4d: add --uninstall-second-opinion (OD-2)


lantisprime:
Deferred open decision OD-2 from the S5 plan (`docs/plans/p4d-s5-s8.md` §17) and confirmed by the S5 code review (finding F6). Filed per Rule 18 step 9.

**Scope:** `install.mjs --uninstall-enforcement` (S5, merged in `8a6dd3b`) deliberately does NOT reverse `--install-second-opinion`. The second-opinion review gate (`second-opinion-gate.mjs` + its PreToolUse registration + the `~/.claude/hooks/second-opinion-providers.json` snapshot) is a separate substrate-adjacent CAPABILITY, not part of the enforcement set, so its removal belongs in its own flag.

**Ask:** add `install.mjs --uninstall-second-opinion` that removes:
- the `second-opinion-gate.mjs` registration (PreToolUse) from settings,
- the deployed gate file,
- the `~/.claude/hooks/second-opinion-providers.json` snapshot (this one IS global — distinct from enforcement which is per-project),

with the same containment + atomic-settings + idempotence guarantees as `--uninstall-enforcement`.

5-field DEFER justification:
1. **Run the scenario:** N/A — feature request, not a bug. S5's E2E never installs second-opinion before uninstalling (correctly, since enforcement uninstall must not touch it).
2. **Spec check:** Plan §5 Non-Goals explicitly excludes this from S5; no spec requires it in P4d S5.
3. **History check:** Reviewer F6 flagged the uncovered scope boundary; intentional capability/enforcement separation.
4. **Same-class check:** Only member of the "uninstall a non-enforcement capability" class today; no siblings left vulnerable.
5. **Residual-risk statement:** Until shipped, an operator who ran `--install-second-opinion` has no scripted uninstall for it (manual removal of the snapshot + registration). No safety impact; the review gate failing closed is the existing safe default.

Source: review reply episode `20260621-065756-…590e` (F6); plan OD-2.

Issue #414, opened 2026-06-21
RFC-008 P4d: uninstall-enforcement version-skew under-removal (review F5)


lantisprime:
Deferred finding from the S5 (`--uninstall-enforcement`) second-opinion code review (claude-subagent, reply episode `20260621-065756-…590e`). Filed per Rule 18 step 9.

**Verdict:** DEFER (out of scope for S5; mirrors install's own readdir-coupling).

5-field DEFER justification:

1. **Run the scenario:** Not reproduced live — no cross-version fixture exists. `enforcementHookLibBasenames(REPO_DIR)` and `enforcementEntryScripts(REPO_DIR)` return the CURRENT repo's basenames. A project installed from an OLDER repo whose `.sh`/`.mjs` was since renamed, then uninstalled with a NEWER repo, would leave the old basename file behind (the new readdir set no longer names it).

2. **Spec check:** No requirement mandates cross-version uninstall. The plan (`docs/plans/p4d-s5-s8.md` §5 Non-Goals) scopes S5 to the current-repo enforcement delta. So no spec forces enforcement here.

3. **History check:** install itself has the identical same-repo readdir-coupling and handles renames only via the explicit `RENAMED_REMOVED` list (`install.mjs` ~1329). The reviewer confirmed this matches existing design.

4. **Same-class check:** The sibling under-removal class also covers `enforcementEntryScripts`/`enforcementBundleLibs`. All share the same current-repo assumption; none are individually more vulnerable. Documented in the `enforcementHookLibBasenames` doc comment (committed in S5).

5. **Residual-risk statement:** If it happens, a stale renamed lib/script lingers under `<project>/.claude/hooks/` after uninstall. It is inert (nothing sources or registers it — the registrations were pruned). No corruption, no fail-open; worst case is a harmless orphan file the operator can delete. Recovery: re-run uninstall from the matching repo version, or delete by hand.

**Possible resolution (when/if cross-version uninstall becomes a requirement):** add a `RENAMED_REMOVED`-style explicit historical-basename list to the uninstall removal set, mirroring install's rename handling.

Source: `scripts/lib/install-manifest.mjs` `enforcementHookLibBasenames` (doc note added in S5 commit `8a6dd3b`).

PR #416, opened 2026-06-21
feat(rfc-008): P4d S5 — install.mjs --uninstall-enforcement [--purge-config]

lantisprime:
## RFC-008 P4d S5 — per-project enforcement uninstall

Adds `install.mjs --uninstall-enforcement [--purge-config]`, reversing the per-project enforcement install delta (RFC-008 P4d, decouple enforcement from the substrate).

### What it does
- Prunes the 9 enforcement hook registrations from `<project>/.claude/settings.json` (canonical exact-match + legacy relative spellings; foreign same-basename hooks are warned and left — REQ-5).
- Deletes the enforcement-only files: the gates + engine + classifier + markers + the `hooks/lib/*.sh` closure, and the enforcement-only `hooks/patterns` + `hooks/plugins` dirs (wholesale — core never creates them).
- **Preserves** the core bp1 set (`bp1-*.mjs` + closure libs + `bp1-approval-check.sh` H1 + `bp1-sweep-on-session.sh` H2), the operator-owned `enforce-config.json` (unless `--purge-config`), and the global substrate (`~/.claude`, `~/.episodic-memory`).

### Correctness model
Primary proof is the single-mock **core-state delta** E2E: in ONE project, core-install → snapshot → `--install-enforcement` → `--uninstall-enforcement` → re-snapshot deep-equals the core baseline (`.claude` tree + parsed settings). It auto-catches over-removal (bp1), under-removal (the `.sh` libs), and residual empty event keys / `patterns`+`plugins` dirs without trusting the hand-derived removal list.

One new manifest export (`enforcementHookLibBasenames`) is readdir-coupled to install's `5_lib` deploy so the deploy and removal sets agree by construction (Rule 14).

### Safety
- `assertContained` (realpath both sides + `path.relative`, not bare `startsWith`) fails closed on a symlinked hook/dir escaping the project tree, including the sibling-prefix case (`.claude/hooks-backup/`).
- Atomic on malformed `settings.json`: parse first; on `SyntaxError` the whole op aborts (no delete, no write).
- Idempotent: nothing-installed run + double run both exit 0 with no change.

### Review (Rule 18)
- Plan reviewed to consensus pre-impl (negative-scenario-planner + second-opinion, 3 rounds, ACCEPT).
- Implementation reviewed via second-opinion (`--provider claude-subagent`): **overall ACCEPT, 0 blocking**. Findings dispositioned — **F1** (operator command bundled in the same `hooks[]` entry as a gate was collaterally deleted by whole-entry removal; reproduced on disk) **fixed inline** with a command-granularity prune + regression test; **F2** `.sh`-lib drift guard added as a co-located machine check; **F3/F6** help-text clarifications; **F4** explicit bp1 H1 assertion; **F5** version-skew → #414; **OD-2** `--uninstall-second-opinion` → #415.

### Tests
- `tests/test-uninstall-enforcement.mjs` — **14/14** (core-state delta, registrations removed, files removed, bp1 preserved, switch preserved, purge, no-global-touch, foreign-hook preserved, idempotent, containment fail-closed, malformed-settings atomic, cwd-independent, F1 shared-entry, F2 drift guard).
- Regression green: contract-deploy 36, lib-closure 6, activation-scoping 13, enforcement-scope 17, s4-gate-e2e 5, bp1-wiring 23.

Follow-ups: #414 (F5 version-skew), #415 (OD-2). OD-3 (branch-protection required-check) lands with S8.

### DIFF
--- a/install.mjs
+++ b/install.mjs
@@ -24,7 +24,7 @@ import { eventsVersion } from './scripts/lib/version-hash.mjs'
 import { findEnforcementTokens } from './scripts/lib/em-recall-purity.mjs'
 import {
   HOOK_SPECS, SESSION_END_SCRIPT, ENFORCEMENT_HOOK_SCRIPTS,
-  enforcementHookFileBasenames, enforcementRegistrations,
+  enforcementHookFileBasenames, enforcementRegistrations, enforcementHookLibBasenames,
   isEnforcementEntryScript, isSubstrateScript, enforcementEntryScripts, enforcementBundleLibs,
   globalScriptLibs, relocatedOnlyLibs, bp1EntryScripts, bp1ClosureLibs,
 } from './scripts/lib/install-manifest.mjs'
@@ -53,6 +53,13 @@ const installHooksForce = argv.includes('--install-hooks-force')
 // registration-green, NOT functional-safe (the registered gates hard-deny on
 // missing deps in a non-this-repo project until S3).
 const installEnforcement = argv.includes('--install-enforcement')
+// RFC-008 P4d S5: --uninstall-enforcement removes a project's enforcement-ONLY
+// set (gates + engine + classifier + markers + the 7 hooks/lib/*.sh + contract
+// config + plugins index + the 9 registrations) while PRESERVING the core bp1
+// set, the operator-owned enforce-config.json, and the global substrate.
+// --purge-config additionally deletes the operator switch (explicit opt-in).
+const uninstallEnforcement = argv.includes('--uninstall-enforcement')
+const purgeConfig = argv.includes('--purge-config')
 const installSecondOpinion = argv.includes('--install-second-opinion')
 const bootstrapLastPrompt = argv.includes('--bootstrap-last-prompt')
 const REPO_HOOKS = path.join(REPO_DIR, 'plugins', 'claude-code', 'hooks')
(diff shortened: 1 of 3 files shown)


lantiscooperdev (commented):
## Code review — RFC-008 P4d S5 (`--uninstall-enforcement`)

Reviewed the removal-set algebra, registration prune, containment, atomicity, and the test coverage against the diff and the live install/uninstall flow.

### Verified sound
- **Removal algebra** — `rootRemove`/`libRemove` = (enforcement sets) MINUS (core bp1 sets), placed in the correct subdir (root vs `lib/`); enforcement-only `patterns/`+`plugins/` removed wholesale (core never creates them). bp1 `.mjs`, `bp1-approval-check.sh`, `bp1-sweep-on-session.sh`, and bp1-shared `.mjs` libs are correctly excluded.
- **Registration prune** — canonical exact-match strings are byte-identical to what install writes (`shellQuote(abs)` for gates, `node <quoted abs>` for SessionEnd); stop-gate's Stop + SubagentStop both handled; legacy-relative spellings resolved + removed; foreign same-basename hooks warned and left.
- **Containment** — `assertContained` realpaths both sides and uses `path.relative`, so symlink-escape and the `.claude/hooks-backup/` sibling-prefix both fail closed with the escape target intact.
- **Atomicity** — parse-first; `SyntaxError` aborts before any delete/write; settings written before file deletion, so a containment throw never orphans registrations against deleted files.
- **No global touch** — all paths rooted at `projectDir`; the before/after global snapshot is byte-identical.

### Findings dispositioned (from the second-opinion claude-subagent pass)
- **F1** (medium): whole-entry `removeHookEntryByCommand` would collaterally delete an operator command bundled in the same `hooks[]` entry as a gate (reproduced on disk). **Fixed inline** here via the command-granularity `removeHookCommandFromEvent` (drops an entry only when it empties it), covered by `t_operator_command_in_shared_entry_preserved`. Good resolution — keeps the shared install-time helper's whole-entry semantics intact.
- **F2** → co-located drift guard `t_sh_libs_not_sourced_by_core` (Rule 14). **F4** → explicit bp1 H1 assertion. **F3/F6** → help-text. **F5/OD-2** → #414 / #415.

### Tests
`test-uninstall-enforcement.mjs` 14/14 incl. the `core+enforce+uninstall ≡ core` delta; regression suites green (contract-deploy/lib-closure/activation-scoping/enforcement-scope/s4-gate-e2e/bp1-wiring).

No blocking issues. The core-state delta E2E is the right correctness anchor and it holds. Recommend merge after maintainer approval.

lantiscooperdev (approved):

PR #417, opened 2026-06-21
feat(rfc-008): P4d S6 seed-identity coupling guard (REQ-7)

lantisprime:
## RFC-008 P4d S6: seed-identity coupling guard (REQ-7)

Closes the S6 slice of `docs/plans/p4d-s5-s8.md`. Parents: R5 (per-project enforcement switch), R3 (effective-tier).

### What

A coupling guard so the install seed can never silently diverge from `loadEnforceConfig`'s absent-file identity. No data migration and no default change (both still resolve to enforce-ON); the guard turns a future divergence into a loud test failure rather than a silent one (Rule 14).

- `scripts/lib/install-manifest.mjs`: new `ENFORCE_CONFIG_SEED` single-source const (the exact bytes install seeds).
- `install.mjs`: imports and writes the const at the `5b-ec-cfg` seed block (was an inline literal); comment updated.
- `tests/test-enforce-config-seed-identity.mjs` (new): the guard.
- `tests/test-s4-gate-e2e.mjs`: the `active:true` restore step now binds the const (was a second hand-typed copy; review F3).

### How the guard works (semantic, not byte)

The seed bytes `{active:true}` can never byte-equal the object `{active:true,bps:{}}`, so the compare is on the load-bearing `active` field (F4/BL-2). Three composing assertions rule out a false pass:

- a schema-valid pin proves the seed is honor-able (would not hit a fail-closed branch of `loadEnforceConfig`);
- the negative control (`active:false`) proves `loadEnforceConfig` actually reads `active` on the honored branch;
- the primary asserts the seed's `active` equals the identity's `active`.

### Verification

| Test | Result |
| --- | --- |
| test-enforce-config-seed-identity.mjs (new guard) | 6/6 |
| test-s4-gate-e2e.mjs (real install + real deployed stop-gate) | 5/5 |
| test-install-contract-deploy.mjs (T4 seed content + schema) | 36/36 |
| test-enforce-config.mjs (loadEnforceConfig regression) | 27/27 |
| test-uninstall-enforcement.mjs | 14/14 |
| test-enforcement-scope.mjs | 17/17 |
| test-activation-scoping-e2e.mjs | 13/13 |
| test-p12-global-clean.mjs | 6/6 |

Rule-14 check: a repo-wide grep confirms the only remaining copy of the seed bytes is the const definition.

### Review (Rule 18 step 6)

Adversarial 3-lens review (test false-pass / Rule-14 drift / REQ-7 conformance), each finding independently verified. 0 blockers.

- F1 (nit): comment overstated the schema-valid pin. Fixed inline (now credits the negative control).
- F3 (low): a second hand-typed seed copy in the s4 test. Fixed inline (binds the const).
- F2 (low): the guard is not wired into CI until S8 (by design; S8 step 8.1 already names this test in the P12 EXPLICIT set). No new issue.

### Out of scope

CI wiring of this guard is S8 (hard dep S6 then S8). No data migration of existing seeded files; no default change.

### DIFF
--- a/install.mjs
+++ b/install.mjs
@@ -23,7 +23,7 @@ import { fileURLToPath } from 'url'
 import { eventsVersion } from './scripts/lib/version-hash.mjs'
 import { findEnforcementTokens } from './scripts/lib/em-recall-purity.mjs'
 import {
-  HOOK_SPECS, SESSION_END_SCRIPT, ENFORCEMENT_HOOK_SCRIPTS,
+  HOOK_SPECS, SESSION_END_SCRIPT, ENFORCEMENT_HOOK_SCRIPTS, ENFORCE_CONFIG_SEED,
   enforcementHookFileBasenames, enforcementRegistrations, enforcementHookLibBasenames,
   isEnforcementEntryScript, isSubstrateScript, enforcementEntryScripts, enforcementBundleLibs,
   globalScriptLibs, relocatedOnlyLibs, bp1EntryScripts, bp1ClosureLibs,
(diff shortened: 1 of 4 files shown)


lantiscooperdev (commented):
## Review verdict: ACCEPT (0 blockers)

Reviewed the S6 seed-identity coupling guard across three layers (per-artifact, cross-file, PR-level) plus an adversarial 3-lens pass (test false-pass, Rule-14 drift, REQ-7 conformance) with independent verification of each finding.

### What I checked

- **Guard correctness / false-pass.** The compare is semantic on `active` (correct: the seed bytes `{active:true}` cannot byte-equal `{active:true,bps:{}}`). The guard is sound: the schema-valid pin establishes honor-ability, and the negative control (`active:false`) proves `loadEnforceConfig` actually reads `active` on the honored branch, closing the "both fell through to identity" false-pass class. Reproduced against an always-fail-closed mutant loader: only the negative control catches that, and it ships in the suite.
- **Rule-14 single-source.** `ENFORCE_CONFIG_SEED` is now the sole source of the seed bytes; `install.mjs` binds it. A repo-wide grep confirms no other hand-copy survives after the s4 test was bound to the const.
- **REQ-7 conformance.** Semantic `active` compare, normalize/ignore `bps`, NEW seeds only, no migration, no default change. Matches plan section 14 Group 2.
- **Cross-platform / wiring.** `os.tmpdir` + `path`; no GNU-isms. Import wired correctly; the install seed write is byte-unchanged.

### Findings (all non-blocking, dispositioned)

- F1 (nit): the original comment overstated the schema-valid pin. Fixed inline (now credits the negative control).
- F3 (low): `test-s4-gate-e2e.mjs` held a second hand-typed seed copy. Fixed inline (binds the const).
- F2 (low): the guard is not in CI until S8 (by design; S8 step 8.1 already names it in the P12 EXPLICIT set).

### Tests (all green)

test-enforce-config-seed-identity.mjs 6/6; s4-gate-e2e 5/5; install-contract-deploy 36/36; enforce-config 27/27; uninstall 14/14; enforcement-scope 17/17; activation-scoping 13/13; p12-global-clean 6/6.

No blockers. User approval required to merge (Rule 17).

lantiscooperdev (approved):

PR #424, opened 2026-06-27
RFC-008 P5: OpenCode enforcement plugin (R6, R10)

lantisprime:
## RFC-008 P5 — OpenCode enforcement plugin (R6, R10)

Adds the OpenCode enforcement plugin: a TypeScript adapter + node bridge that gate repo-source writes per the shared `enforce-contract.mjs` thin waist, deployed per-project (Principle 12), with install/uninstall, CI, and a runbook. Maps to RFC-008 R6 (per-tool plugins) and R10 (enforcement runbooks).

### What ships (slices S1-S6)
- S1: gauntlet parameterized by harness (REQ-1).
- S2: declarative opencode plugin (manifest, `_index` entry, event translations).
- S3: enforcement runbook (§1-§10, agent manifest).
- S4: shared repo-source carve-outs, bash + node at parity (REQ-7).
- S5: TypeScript adapter (`enforcement.ts`) + node bridge (`enforce-bridge.mjs`).
- S6: `install.mjs --tool opencode --install-enforcement` + CI gating the opencode suites (REQ-11/REQ-12). Deploys the contract set project-local beside the engine so the kill switch and project-local resolution work in a clean deploy (R5/P12).

### Pre-PR whole-branch review fixes (last two commits)
A PR-level review of the full `main..HEAD` diff (the layer the per-slice reviews could not provide) found issues no single-slice review saw:

- **Adapter was inert (failed open).** `enforcement.ts` returned a nested hooks object, but the installed `@opencode-ai/plugin` `interface Hooks` uses flat dotted keys (`"tool.execute.before"`) with two-arg `(input, output)` signatures. The host found `undefined` and the only STRONG blocking hook never fired. Rewritten to the installed shape; args read from `output.args`/`input.args`, sessionID from `input.sessionID`, stop via the `session.idle` event.
- **No test loaded the adapter.** Every test drove the bridge directly, so the broken adapter passed every green suite. Added `test-opencode-adapter-conformance.mjs` that loads the deployed `.ts` and drives the real `tool.execute.before` hook to a block on a gated write / allow on a carve-out; it fails loud (not skip) if the runtime cannot strip TS. `plugin-validate.yml` pinned to Node 24 for that step.
[...]

PR #427, opened 2026-06-28
fix(opencode): spawn real node for the bridge, not process.execPath (Bun-runtime fail-close)

lantisprime:
## What

Fix the opencode enforcement adapter fail-closing on **every** gated tool call in the real opencode runtime.

## The bug

`plugins/opencode/capabilities/enforcement.ts` spawned the node bridge with `spawnSync(process.execPath, [BRIDGE])`. Under opencode's Bun-compiled runtime, `process.execPath` is the **opencode binary**, not node. opencode treats the `.mjs` path as a project directory, fails to `cd` into it, exits 0 with **empty stdout** → the adapter's `JSON.parse` throws → `"opencode-enforce: fail-closed: bridge stdout not valid JSON"` on every gated tool call. The enforcement never actually decides; it blocks everything (the agent cannot run bash or write files at all).

## Why it shipped green (#424)

Every existing test is a proxy that runs under **node**, where `process.execPath` IS node:
- `test-opencode-adapter-conformance.mjs` imports the adapter under node.
- `test-opencode-adapter.mjs` drives the bridge directly under node.

The bug only manifests in the real opencode/Bun runtime — the recurring "real-runtime E2E was actually a bridge proxy" gap.

## Root cause (empirically reproduced)

Driving the real opencode binary against the deployed bridge:

```
spawnSync("~/.opencode/bin/opencode", [enforce-bridge.mjs], {input})
  → status: 0, stdout: "",
    stderr: "Error: Failed to change directory to .../enforce-bridge.mjs"
```

## Fix

`resolveNodeExe()` — use `process.execPath` only when its basename is `node`; otherwise locate `node` on PATH (`which`/`where`), with a `"node"` last resort. Spawn the resolved node for the bridge.

## Regression test (actual terminal, not proxy)

`tests/test-opencode-enforcement-live-e2e.mjs` drives the **actual opencode binary** via tmux against a mock project with enforcement installed, then asserts the hook decides (no fail-close) and the write lands. It guards the exact `fail-closed|not valid JSON` signature, so it goes red if anyone reverts to `process.execPath`.

Observed both states this change:
- **Before fix:** live opencode agent → `fail-closed` → write blocked.
- **After fix:** hook decides (allows `echo`, read_only) → opencode's own permission prompt → write lands (`src/x.mjs` = `BYPASS_OK`). Automated test: `PASS (noFailClosed=true, wrote=true)`.

`UNGUARDED-IN-CI` (needs the real binary + a reachable model + a TTY).

## Deploy

Post-merge, opencode-enforcement projects need a re-install to pick up the fixed adapter:
`node install.mjs --tool opencode --install-enforcement --project <dir>`.

### DIFF
--- a/plugins/opencode/capabilities/enforcement.ts
+++ b/plugins/opencode/capabilities/enforcement.ts
@@ -33,12 +33,39 @@
 // @ts-ignore — type-only; not compiled in this repo (zero-dep policy for test harness)
 import type { Plugin } from "@opencode-ai/plugin";
 import { spawnSync } from "node:child_process";
-import { realpathSync } from "node:fs";
+import { realpathSync, existsSync } from "node:fs";
 import { fileURLToPath } from "node:url";
-import { dirname, resolve } from "node:path";
+import { dirname, resolve, basename } from "node:path";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const BRIDGE = resolve(__dirname, "enforce-bridge.mjs");
+
+// Resolve a real `node` to run the .mjs bridge. CRITICAL: under opencode's
+// Bun-compiled runtime, process.execPath is the OPENCODE binary, NOT node — so
+// spawnSync(process.execPath, [BRIDGE]) makes opencode treat the .mjs path as a
+// project dir, fail to cd, exit 0 with EMPTY stdout, and the JSON.parse below
+// throws -> fail-closed on EVERY gated tool call (the bridge never runs). Use
+// process.execPath only when it IS node; otherwise locate node on PATH. Bug
+// reproduced 2026-06-28 by driving the live opencode TUI (the bridge-proxy /
+// direct-node tests masked it). Regression: tests/test-opencode-node-resolve.mjs.
+// Exported + parameterized for the regression test: a node-run test can't change
+// its own process.execPath, so it passes a simulated non-node exe to prove the
+// resolution. The default (no-arg) path is cached per process.
+let _nodeExe: string | null = null;
+export function resolveNodeExe(exe: string = process.execPath): string {
+  const useCache = exe === process.execPath;
+  if (useCache && _nodeExe) return _nodeExe;
+  let result: string;
+  if (/^node(\.exe)?$/i.test(basename(exe))) {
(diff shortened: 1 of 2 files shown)


lantiscooperdev (commented):
## Bot code-review pass

Scope: the 2-file diff — `enforcement.ts` (`resolveNodeExe`) + `tests/test-opencode-enforcement-live-e2e.mjs`.

### Correctness
- **ACCEPT** — `resolveNodeExe` gates on `basename(exe)` matching `^node(\.exe)?$`, else resolves `node` via `which`/`where`, with a `"node"` last resort. This addresses the confirmed root cause (`process.execPath` is the opencode binary, not node, under Bun). The live E2E goes green and the manual before/after (fail-closed → decided + write landed) corroborates.
- The per-process cache only short-circuits the default `process.execPath` path; explicit-arg calls bypass it, which keeps the function testable. Good.

### Residuals (non-blocking, documentation-tier)
1. **PATH dependency:** if `node` is not on opencode's spawn PATH, the `"node"` last-resort ENOENTs → `r.error` → throw → fail-closed. Safe (blocks, never opens), but "enforcement actually decides" now requires node on PATH. Worth a one-line runbook note.
2. **Version skew:** `which node` takes the first PATH node, which may differ from a project's intended toolchain node. Low risk for a zero-dep stdlib bridge.
3. **Test approval keypress:** the E2E sends `Enter` assuming "Allow once" is the default-highlighted permission option. If opencode reorders that menu, the keypress could land elsewhere — but the `wrote=true` assertion fails-closed on a wrong approval, so the test stays sound.

### Tests
- The **actual-terminal** E2E is the correct call: the existing `test-opencode-adapter-conformance.mjs` / `-adapter.mjs` run under node and structurally cannot reproduce this class. SKIP-on-missing-precondition (no binary / model / tmux) is honest, not a false pass.

**Verdict: ACCEPT.** Real shipped bug, correct minimal fix, real-runtime regression guard. Residuals are doc-tier, not blockers. User approval in the UI satisfies the required-review gate (Rule 17).

lantiscooperdev (approved):

PR #447, opened 2026-07-04
fix(em-list,em-search): coerce date/time sort keys against malformed index rows

lantisprime:
## Summary

A hand-authored episode in the pi-extensions local store (`20260702-141607-p5b-1-s2-b1bc`, frontmatter used `created:` instead of `date:`/`time:`, bypassing em-store) was appended to `index.jsonl` without `date`/`time` fields. At the date-based sort sites `undefined + undefined = NaN`, and `NaN.localeCompare` throws:

```
TypeError: (b.date + b.time).localeCompare is not a function   (em-list.mjs:64)
```

Reported 2026-07-04 from the Pi Agent harness; em_list crashed on every invocation while em_search appeared fine (default scoring path sorts numerically; the same crash is reachable via `--no-score` at em-search.mjs:297).

## Fix

Template-literal coercion with `?? ''` at the two reachable sites (`em-list.mjs:64`, `em-search.mjs:297`). Malformed rows sort last in the descending order instead of crashing.

Complete class analysis (all `.localeCompare` sort sites over index fields):
- `em-recall.mjs:235` — NOT changed; filter requires `typeof e.date === 'string'` before the sort.
- `em-check-stale.mjs:79` — NOT changed; `fetched < cutoffStr` rejects undefined/numeric before the sort.

## Verification

- Regression test `tests/test-em-sort-missing-datetime.mjs`: red-then-green. Pre-fix 3/4 fail with the TypeError at the exact target lines; post-fix 4/4 pass. Fixture reproduces both malformed shapes (no date/time; numeric date, no time) via raw JSONL append, exactly how the foreign harness corrupted the store.
- E2E, strong layer: fixed em-list run with cwd = the real pi-extensions store (malformed row present) returns `{"status":"ok","count":10,...}` exit 0. Negative control: unfixed em-list on the same store exits 1 with the TypeError.

## Deferred (issue to follow)

Writer-side hole: the substrate accepts hand-appended index rows missing date/time and `em-rebuild-index` copies frontmatter verbatim without backfill. Tracking issue filed separately and cross-referenced.

### DIFF
--- a/scripts/em-list.mjs
+++ b/scripts/em-list.mjs
@@ -61,10 +61,14 @@ if (project) {
   results = results.filter(e => e.project === project)
 }
 
-// Coerce sort keys: foreign-harness writers have appended index rows without
-// date/time (undefined + undefined = NaN, and NaN has no .localeCompare).
-// Malformed rows sort last in the descending order instead of crashing.
-results.sort((a, b) => `${b.date ?? ''}${b.time ?? ''}`.localeCompare(`${a.date ?? ''}${a.time ?? ''}`))
+// Sort key tolerant of foreign-harness index rows (hand-appended without
+// em-store): only string-typed fields participate. A non-string date (absent,
+// null, numeric) would either crash (NaN.localeCompare) or, stringified,
+// out-sort ISO keys ("20260703" > "2026-07-04..."); time joins only when date
+// is a string so a time-only key like "23:59" cannot beat ISO date keys.
+// Rows without a string date get key '' and sort last in descending order.
+const dtKey = e => typeof e.date === 'string' ? e.date + (typeof e.time === 'string' ? e.time : '') : ''
+results.sort((a, b) => dtKey(b).localeCompare(dtKey(a)))
 results = results.slice(0, limit)
 
 const output = results.map(({ _source, ...rest }) => ({ ...rest, source: _source }))
--- a/scripts/em-list.mjs
+++ b/scripts/em-list.mjs
@@ -61,7 +61,10 @@ if (project) {
   results = results.filter(e => e.project === project)
 }
 
-results.sort((a, b) => (b.date + b.time).localeCompare(a.date + a.time))
+// Coerce sort keys: foreign-harness writers have appended index rows without
+// date/time (undefined + undefined = NaN, and NaN has no .localeCompare).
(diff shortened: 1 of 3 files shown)


lantiscooperdev (commented):
Substantive review (bot account, per Rule 17; approval stays with the repo owner).

## Scope and correctness

- Root cause verified against the real corruption: pi-extensions index row `20260702-141607-p5b-1-s2-b1bc` has no `date`/`time` (hand-authored, bypassed em-store). Pre-fix `em-list` crashes on it (`TypeError: (b.date + b.time).localeCompare is not a function`, em-list.mjs:64); post-fix run against that same store returns `{"status":"ok","count":10}`.
- The class enumeration is complete for read-side `.localeCompare` sort sites: em-list.mjs (always reachable) and em-search.mjs (reachable under `--no-score`) are fixed; em-recall.mjs:235 and em-check-stale.mjs:79 are correctly left unchanged because their membership filters already exclude non-string keys.
- Round-2 policy (`dtKey`: only string-typed fields participate, `time` only when `date` is a string) is the right call over plain stringification. Stringified numeric dates ("20260703") out-sort ISO keys and let `--limit` evict well-formed episodes; independent codex review round 1 caught this (REJECT), round 2 re-verified at c43a1a9 and returned ACCEPT with no findings, including its own shell repros for the limit interaction, nested-cwd, and linked-worktree store resolution.

## Tests

- `tests/test-em-sort-missing-datetime.mjs` follows the sandbox conventions (mkdtemp HOME, real em-store seeding, spawnSync of the real scripts, assertions on captured stdout/exit codes). Matrix covers absent/absent, number/absent, number/number, null/null, and absent-date/string-time ("23:59", the key that would beat ISO dates if time participated alone). Sort-last is asserted per malformed row in both scripts, plus the `--limit 1` eviction case.
- Red-then-green is documented: 3/4 failed pre-fix at the exact target lines; the round-2 assertions' red is the codex round-1 captured repro output.

## Deferred

Writer-side hole (substrate accepts hand-appended rows; em-rebuild-index copies frontmatter verbatim, no backfill) is tracked in #448 with the 5-field DEFER.

No blockers from this review. CI: p12-invariant-gate + validate x2 all pass.

lantiscooperdev (approved):

PR #449, opened 2026-07-04
feat(scripts): --help/-h short-circuit on all 21 installed substrate scripts

lantisprime:
## Summary

Most substrate scripts had no `--help` handler: an unknown flag was silently ignored and the script ran its default behavior. During the install-guide work (2026-07-04) an agent probing `em-prune --help` executed a **real prune pass** (em-prune is destructive without `--dry-run`), `em-rebuild-index --help` rewrote the index, and `em-seed-patterns --help` would have seeded. Every AI harness probes CLIs with `--help`, so this was a standing footgun. Requested by the operator as an implementation, not an issue.

## Change

Each of the 21 installed substrate scripts (all `em-*.mjs` + `second-opinion.mjs`) now short-circuits on `--help`/`-h` as the first executable logic after `argv`, before any filesystem, store, or subprocess side effect:

```json
{"status":"help","script":"em-prune.mjs","usage":"node em-prune.mjs [--scope local|global|all] [--threshold <n>] [--dry-run] [--check]"}
```

- Usage strings derive from each script's Usage header or its actual argv parser; flag audit confirmed no invented flags (spot-checked `em-watch-codex --limit`, `em-recall --warn-time-ms/--warn-count`).
- `em-restore` already had a handler; aligned its `status` from `ok` to `help`.
- `em-lock` scans only tokens before the `--` separator so a wrapped command's own `--help` is not intercepted.
- No other behavior changed; unknown non-help flags keep their current semantics.

## Verification

- `tests/test-em-help-flags.mjs`: 27/27. All 21 scripts on `--help` (exit 0, JSON `status: help`, non-empty usage, **no** `.episodic-memory` store created in the sandbox), the `-h` alias, two discrimination controls (default `em-list` runs; default `em-store` DOES create the store, proving the no-store assertion bites), and belt-and-braces checks that prune/seed/rebuild emit no work keys on `--help`.
- Regression suites green: tag-flags 11/11, sort-missing-datetime 5/5, body-file 20/20, revise-scope-inherit 18/18, watch-codex 25/25, restore self-test 120/120, seed-patterns 15/15, mine-transcripts 9/9, audit-compliance 7/7.
- Live E2E of the original footgun: `em-prune --help` against the real environment prints the help JSON and performs no prune.

### DIFF
--- a/scripts/em-audit-compliance.mjs
+++ b/scripts/em-audit-compliance.mjs
@@ -44,6 +44,12 @@ import { walkTranscripts, groupBySession } from './lib/transcript-walker.mjs'
 // ---------------------------------------------------------------------------
 
 const argv = process.argv.slice(2)
+
+if (argv.includes('--help') || argv.includes('-h')) {
+  console.log(JSON.stringify({ status: 'help', script: 'em-audit-compliance.mjs', usage: 'node em-audit-compliance.mjs [--since <ISO>] [--slug <substring>] [--exclude-worktrees] [--format json|markdown] [--prior-since <ISO>]' }))
+  process.exit(0)
+}
+
 function flag(name, def = undefined) {
   const i = argv.indexOf(name)
   if (i === -1) return def
--- a/scripts/em-backup.mjs
+++ b/scripts/em-backup.mjs
@@ -1675,6 +1675,11 @@ function selfTest() {
 // ---------------------------------------------------------------------------
 const argv = process.argv.slice(2)
 
+if (argv.includes('--help') || argv.includes('-h')) {
+  console.log(JSON.stringify({ status: 'help', script: 'em-backup.mjs', usage: 'node em-backup.mjs (--audit | --init | --sync | --self-test | --show-config) [--sample <n>]' }))
+  process.exit(0)
+}
+
 // Codex PR-#137 round-5: every CLI JSON output (success OR error) must pass
 // through artifact redaction at the output boundary. Patching individual
 // return objects + error strings was whack-a-mole. This is the LAST line of
--- a/scripts/em-check-stale.mjs
+++ b/scripts/em-check-stale.mjs
@@ -21,6 +21,11 @@ const LOCAL_DIR = resolveLocalDir()
 
 const argv = process.argv.slice(2)
 
+if (argv.includes('--help') || argv.includes('-h')) {
+  console.log(JSON.stringify({ status: 'help', script: 'em-check-stale.mjs', usage: 'node em-check-stale.mjs [--days <n>] [--project <name>] [--scope local|global|all]' }))
+  process.exit(0)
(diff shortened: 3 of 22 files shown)


lantiscooperdev (commented):
Substantive review (bot account, per Rule 17; approval stays with the repo owner).

## Scope and correctness

- Real footgun, verified live: before this PR, `em-prune --help` executed an actual prune pass (`dryRun = argv.includes('--dry-run')`, so bare runs are destructive), `em-rebuild-index --help` rewrote the index, and most readers silently ran defaults on unknown flags. Observed in practice on 2026-07-04 when an agent probed the scripts while drafting the install guides.
- All 21 installed substrate scripts now short-circuit on a standalone `--help`/`-h` before any filesystem, store, or subprocess side effect, emitting exactly one JSON object `{status:'help', script, usage}` with exit 0. Handler placement audited per-file; only em-restore's sits inside its existing CLI dispatcher (pre-existing structure, side-effect-free on the help path).
- Usage strings audited against the real parsers (spot checks: `em-watch-codex --limit`, `em-recall --warn-time-ms/--warn-count` all exist); no invented flags.
- `em-lock` deliberately scans only tokens before its `--` separator so a wrapped command owns its own `--help`; pinned by test.

## Review history (three codex rounds, independent model)

- R1 REJECT: em-restore help lacked the `script` key; test assertions too loose to catch it; value-position `--help` (e.g. `--summary --help`) changes behavior vs strict byte-for-byte compatibility.
- R2: B1/B2 verified fixed; the narrowed token semantics (standalone `--help` triggers help anywhere, pinned by a regression test showing `em-store --summary --help` stores nothing) explicitly accepted; one remaining P2, the em-lock test oracle passed vacuously because `node -e ... --help` printed node's own help.
- R3 ACCEPT, no findings, verified at commit 18a8770 with its own sandbox runs (29/29).

The strengthened exact-contract assertion earned its keep immediately: it caught em-restore's pretty-printed multi-line output on first run, exactly the class R1-B2 predicted.

## Tests

`tests/test-em-help-flags.mjs`, 29 assertions: exact contract per script (exit 0, empty stderr, single JSON value, keys exactly {script,status,usage}, script === basename, no store created in the sandbox), `-h` alias, pinned value-position semantics, em-lock separator exception with a positive wrapped-ran oracle, two discrimination controls, and belt-and-braces checks that prune/seed/rebuild emit no work keys. Regression suites reported green across the em-* surface (tag-flags 11/11, sort 5/5, body-file 20/20, revise-scope 18/18, watch-codex 25/25, restore 120/120, seed-patterns 15/15, mine-transcripts 9/9, audit-compliance 7/7).

No blockers from this review. CI: p12-invariant-gate + validate jobs all pass.

lantiscooperdev (approved):

PR #470, opened 2026-07-08
Fix #469: null-proto index maps so "constructor" keys can't hit Object.prototype

lantisprime:
Closes #469.

## Bug

Index maps keyed by episode-derived strings (tokens, tags, category names, stats tallies) were plain `{}` objects. A key colliding with an `Object.prototype` property name — `constructor` is the only one that survives the `[^a-z0-9]+` tokenizer — resolved to the inherited `Object` function instead of `undefined`.

Confirmed failures (all reproduced at runtime, not traced):
- `em-store` / `em-revise`: crash on storing/revising any episode whose text contains the word `constructor` (the issue's live global-store edge)
- `em-rebuild-index`: crash on any store containing the token, plus the literal unknown-category key path
- `em-search --tag constructor`: "not iterable" crash
- `em-move`: **partial move** — episode+index moved, destination `tags.json` update crashed (codex round-1 repro)
- `em-restore`: crash mid-apply restoring a backup containing the tag (codex round-1 repro)
- `em-seed-patterns`: silent misroute — inherited `Object.prototype.constructor` read as "already seeded" (`seeded:0, skipped:1`), and crash flushing the tag (codex round-1 repro)
- `em-stats` / drift tallies: garbage string counters (corruption, not crash)

## Fix

One idiom everywhere: fresh maps are `Object.create(null)`; `JSON.parse`-loaded maps are rehydrated via the new `nullProtoIndex()` helper in `scripts/lib/relevance.mjs` (loaders + all incremental updaters). `em-prune`'s local loader and the `parseFrontmatter`/redact-clone maps are untouched — they are `Object.keys`-walked or assignment-only, which never sees inherited properties (verified by codex sweep, round 2).

## Tests (Rule 15)

`tests/test-prototype-key-collision.mjs` — 12 runtime probes on isolated fixture stores: store/search(tag+query)/revise/rebuild/stats/move/seed with `constructor` as token, tag, `pattern_id`, and hand-authored category, plus the isolated-mechanism units. Negative control: the suite's scenario run against the pre-fix scripts reproduces `TypeError: index[tok].includes is not a function` at `relevance.mjs:267`.

Wired into CI (`plan-marker-validate.yml`), along with `test-recall-v2.mjs` which was previously not run by any workflow — that gap is why no CI fixture ever contained the colliding token.

## Second opinion

Codex (harness, provider codex): round 1 HOLD — F1 found the `em-move`/`em-restore`/`em-seed-patterns` writers my sweep missed, each with a runtime repro. ACCEPTED and fixed in the round-2 commit. Round 2 ACCEPT — codex reran its own round-1 repros clean and confirmed no remaining unsafe inverted-index/tally writers on episode-derived keys. Replies: `.review-store/replies/20260708-021416-a8f576a0dbef.body.md`, `.review-store/replies/20260708-022553-4d9de2c2bd4e.body.md`.

## Post-merge deploy note

Global scripts at `~/.episodic-memory/scripts/` currently carry the bug (deployed from pre-fix main). After merge: redeploy via `install.mjs`, then `em-rebuild-index --scope all` should succeed on the global store that crashes today.

### DIFF
--- a/.github/workflows/plan-marker-validate.yml
+++ b/.github/workflows/plan-marker-validate.yml
@@ -192,6 +192,12 @@ jobs:
       - name: Run RFC-009 P1a categories.json global-deploy mock E2E
         run: node tests/test-install-categories.mjs
 
+      - name: Run issue #469 prototype-key-collision regression tests
+        run: node tests/test-prototype-key-collision.mjs
+
+      - name: Run recall-v2 token-index suite (tokens.json write/read parity)
+        run: node tests/test-recall-v2.mjs
+
   p12-invariant-gate:
     # RFC-008 P4d S8 (REQ-10 / REQ-11): the three Principle-12 invariants as ONE
     # separately-named REQUIRED check. Membership is defined in code
--- a/scripts/em-move.mjs
+++ b/scripts/em-move.mjs
@@ -43,7 +43,7 @@ import crypto from 'crypto'
 import { execFileSync } from 'child_process'
 import { fileURLToPath } from 'url'
 import { resolveLocalDir } from './lib/local-dir.mjs'
-import { normalizeTags, episodeTokens, updateTokensIndex } from './lib/relevance.mjs'
+import { normalizeTags, episodeTokens, updateTokensIndex, nullProtoIndex } from './lib/relevance.mjs'
 import { canonicalCategory } from './lib/categories.mjs'
 
 const GLOBAL_DIR = path.join(os.homedir(), '.episodic-memory')
@@ -131,8 +131,10 @@ function removeFromInverted(dataDir, fileName, id, warnings, pretty) {
 function addToInverted(dataDir, fileName, id, keys, pretty) {
   if (keys.length === 0) return
   const p = path.join(dataDir, fileName)
-  let idx = {}
-  try { idx = JSON.parse(fs.readFileSync(p, 'utf8')) } catch {}
+  // Null-proto: a tag/category key named "constructor" must not resolve to
+  // Object.prototype (issue #469)
+  let idx = Object.create(null)
+  try { idx = nullProtoIndex(JSON.parse(fs.readFileSync(p, 'utf8'))) } catch {}
   for (const key of keys) {
     if (!idx[key]) idx[key] = []
     if (!idx[key].includes(id)) idx[key].push(id)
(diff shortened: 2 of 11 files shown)


lantiscooperdev (commented):
Substantive review (bot account, Rule 17 — comment only; user approval remains the gate).

**Scope**: commits 59d4270 + 4de8404 against issue #469 — prototype-key collision in index maps.

**Mechanism check**: the fix is the correct minimal class fix. `Object.create(null)` for fresh maps and `nullProtoIndex()` rehydration after every `JSON.parse` remove the entire `Object.prototype` lookup surface, rather than special-casing the `constructor` key. The existing truthiness guards (`if (!index[k])`, `||=`) become sound once the prototype is null, so the diff stays idiom-preserving and small.

**Class completeness**: verified by two independent sweeps (author grep + codex adversarial round 1). Round 1 caught three writers the first sweep missed (`em-move` addToInverted — partial-move data-loss path; `em-restore` tags/category merge; `em-seed-patterns` seeded-check silent misroute + flush crash); all fixed in 4de8404 with the same idiom. Remaining plain `{}` maps on episode-derived keys are assignment-only field bags or `Object.keys`-walked (safe — inherited props are never own-enumerable), confirmed in codex round 2.

**Evidence quality**: all failure claims are runtime-reproduced, not traced. Negative control demonstrates the exact pre-fix crash (`TypeError: index[tok].includes is not a function`, relevance.mjs:267 via em-store.mjs:214). FP control: a legitimately-stored own `constructor` key round-trips store → rebuild → search. JSON byte-compat of null-proto maps vs plain maps verified (`same:true`, codex round 1).

**Tests/CI**: regression suite covers write path, read path, rebuild, stats, move, and seeding, and is wired into `plan-marker-validate.yml` (runs on every PR, no path filter). Wiring the previously-orphaned `test-recall-v2.mjs` into CI closes the exact gap that let this ship: the tokens.json suite existed but never ran.

**Residual risk**: low. The deployed global scripts still carry the bug until post-merge redeploy; the PR body documents the redeploy step. Second-level nesting is a non-issue today (all persisted index values are arrays, verified), but any future nested-map index format must apply `nullProtoIndex` per level — worth remembering if tokens.json ever gains structure.

Verdict: no blocking findings from this review. Codex second-opinion round 2: ACCEPT.

lantiscooperdev (approved):

PR #572, opened 2026-07-23
add contradiction advisory for conflicting decisions (#537)

lantisprime:
Closes #537 (parts 1 and 2; part 3 deferred to #571).

## The problem

From #537's isolated-fixture repro: store decision A ("Config files use JSON format"), then a directly contradicting decision B ("Config files use YAML format") with plain `em-store`. B returns `{"status":"ok"}` with **no warning of any kind**, both stay `active`, both come back from search, and `em-doctor` reports 16 ok / 0 warn because it audits structure and never semantics.

## What this adds

An advisory safety net at both layers, sharing one comparator. Advisory throughout: nothing blocks, nothing gates, no exit code changes.

**`scripts/lib/contradiction.mjs` (new).** Compares index rows by summary-token Jaccard at `SUMMARY_JACCARD_MIN = 0.6` — deliberately 3x the repo's existing clustering precedent (`topic-tracks/config.json` uses 0.2), because clustering wants recall and an advisory wants precision. The canonical repro scores exactly 0.667. Four exclusion axes: different project, non-`decision` category, non-`active` status, and pairs already linked transitively by `supersedes` in either direction. An empty token set matches nothing, including another empty set — an empty summary must not compare equal to itself.

**`em-store` write-time advisory.** For a `decision` write, prints one informational stderr line per similar active decision plus an `em-revise` hint. Placed after the write and after lock release, inside `try {} catch {}`, directly mirroring the existing R9a trigger-collision report at `em-store.mjs:408-427`. Stdout stays exactly `{status,id,file,scope}` and exit stays 0 — asserted by test while the advisory is firing.

**`em-doctor` audit-time check.** New `contradiction-candidates` store-class check between `supersedes-links` and `tmp-litter`, reusing the index rows `checkStore` already loads (no new file reads — `em-doctor` stays index-only and never opens an episode body). `warn` on candidates, `ok` otherwise, no `fix:` hint because the repair is human judgment. Project groups above `MAX_GROUP` (2000) are skipped **and named** in `skipped_projects`, never silently truncated.

## Not in scope

The RFC-001 scoring contract is byte-unchanged (`git diff --stat main -- scripts/lib/relevance.mjs` → empty). #537's own constraints require this: "detection is a separate advisory pass, never a re-ranking change." The `access_count` entrenchment sub-item is deferred to #571 with all five DEFER fields.

## Verification

All run against the branch; runtime probes drive the **real** scripts against isolated fixture stores (`mkdtemp` cwd + `mkdtemp` HOME).

| Check | Observed |
|---|---|
| `node tests/test-contradiction.mjs` | **20 passed, 0 failed** |
| `node tests/test-contradiction.mjs --break-detector` | **exit 1, 4 failed** — guard proven RED |
| `node tests/test-em-doctor.mjs` | 13 passed, 0 failed |
| `node tests/test-relevance.mjs` | 17 passed, 0 failed |
| `node tests/test-store-write-concurrency.mjs` | 46 passed, 0 failed |
| `node tests/test-ci-suite-registration.mjs` | 16 passed, 0 failed |
| `git diff --stat main -- scripts/lib/relevance.mjs` | empty |

8 of the 20 tests are end-to-end: real `em-store` ×2 → assert the advisory names the first episode's id and that stdout is still exactly four keys with exit 0; real `em-doctor` → assert the `warn` row with `data_dir`, that `--strict` flips exit to 1, and that a clean store reports `ok`; real `em-revise` chain → assert the sanctioned correction reports `ok` and does **not** false-positive. Both polarities are covered at every call site.

`--break-detector` raises the pure-layer threshold to an unreachable 1.1, so the positive assertions must go red. A guard never observed failing guards nothing.

## Review

Reviewed by a read-only pi/GLM-5.2 frozen-diff seat against the on-disk spec (`docs/plans/issue-537-contradiction-advisory.md`) as ground truth, in three layers (per-artifact, cross-file, whole-diff) with ten behavior-simulation probes.

**Verdict: ACCEPT, 0 blocking findings.**

Nine non-blocking findings, all dispositioned in the plan's §21:

- Fixed in the fold round: cycle-guard test fidelity (the original test used a linear chain, so the `seen` guard was never the cause of termination), the `MAX_GROUP` default not being pinned (a regression to `Infinity` would have passed), `data_dir` carriage not asserted, and a spec export-count typo. These added `testChainCycleTerminates` and `testDefaultMaxGroupIsPinned`.
- Deferred with full 5-field justifications: #568 (negation and short-subset false positives), #569 (non-ASCII summaries tokenize to empty — the fix must change `lib/relevance.mjs`, which this slice forbids), #570 (single-scope chain resolution and strict category equality; the latter verified unreachable today — `grep -c deprecated categories.json` → 0).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -344,6 +344,9 @@ jobs:
       - name: Run relevance suite (wave 1)
         run: node tests/test-relevance.mjs
 
+      - name: Run contradiction advisory suite (#537)
+        run: node tests/test-contradiction.mjs
+
       - name: Run em-move suite (RFC-005)
         run: node tests/test-em-move.mjs
 
--- a/docs/plans/issue-537-contradiction-advisory.md
+++ b/docs/plans/issue-537-contradiction-advisory.md
@@ -1,0 +1,1310 @@
+# ISSUE-537 Contradiction / near-duplicate advisory Plan
+
+> **FROZEN BUILD SPEC (playbook v5).** This document is the seat contract. The builder
+> brief points here and adds nothing. The reviewer's ground truth is this same document:
+> every ACCEPT / HOLD finding cites a section of it. The spec is frozen at dispatch —
+> a mid-build scope change requires the orchestrator to edit this file AND re-brief the
+> seat explicitly.
+
+## §1 Status
+
+`Approved — build.` Executor altitude **LOW** (§0.1): the implementer is a pi/MiniMax-M3
+builder seat with fresh context. **Appendix A is the build path.**
+
+| Field | Value |
+|---|---|
+| RFC | `RFC-001` (scoring contract — constraint only, not modified) |
+| Parent requirements | RFC-001 Phase 2 lines 75-79 / 81-88 (constraint); `CAPABILITIES.md:85-92` curation family; `PRINCIPLES.md` P1, P6 |
+| GitHub issue | `#537` |
+| Target branch | `feat/issue-537-contradiction-advisory` |
+| Executor altitude (§0.1) | **low** |
+| Worktree | `/Users/charltondho/Developer/projects/episodic-memory-537` |
+
+## §2 Episode / scout search summary
+
+Scout maps produced 2026-07-23 by four parallel pi/MiniMax-M3 read-only seats; every
+file:line below was re-verified by the orchestrator against the worktree before freezing.
+
+Load-bearing facts (all orchestrator-verified, not scout-asserted):
+
+- `scripts/em-store.mjs:408-427` — an existing **stderr-only, best-effort, post-lock,
+  informational advisory** (the R9a trigger-collision report). This is the **style anchor**
+  for the new advisory: same channel, same placement, same `try {} catch {}` posture.
+  Verified by reading `em-store.mjs:396-430`.
+- `scripts/em-store.mjs:429` — `console.log(JSON.stringify(successPayload))`, the single
+  success-stdout call. Its contract is exactly `{ status, id, file, scope }`
+  (`em-store.mjs:403`). **The advisory must not change it.**
+- `scripts/em-doctor.mjs:123-305` — `checkStore(dataDir, scopeName)`; index rows are loaded
+  once at `:168` into `entriesForShape` (aliased `entries` at `:183`). Verified by reading
+  `em-doctor.mjs:123-222` and `:222-316`.
+- `scripts/em-doctor.mjs:257` — `// --- tmp-litter ---...` is the insertion anchor: the new
+  check goes between `supersedes-links` (ends `:255`) and `tmp-litter`.
+- `scripts/em-doctor.mjs:60` — `loadIndex` is already imported from `./lib/relevance.mjs`.
+- `scripts/em-doctor.mjs:563-567` — `checkStoreWithDir` stamps `data_dir` onto every row a
+  store-class check emits, which is why the new check belongs **inside `checkStore`**, not
+  in the bottom run-block.
+- `scripts/em-doctor.mjs:646-656` — summary tally, `healthy = summary.error === 0 && (!strict
+  || summary.warn === 0)`, exit `healthy ? 0 : 1`. A `warn` alone keeps exit 0; `--strict`
+  turns it into exit 1.
+- `scripts/lib/relevance.mjs:213-218` — `tokenizeQuery(text)`: lowercased, split on
+  `[^a-z0-9]+`, length ≥ 2, deduped, order preserved. Verified by reading the function.
+- `scripts/topic-tracks/config.json:3-4` — existing Jaccard precedent in this repo
+  (`tag_jaccard_min` 0.3, `summary_jaccard_min` 0.2, used **OR**-combined for clustering).
+  This plan deliberately uses a **much higher** 0.6, because clustering wants recall and an
+  advisory wants precision.
+- `docs/rfcs/RFC-001-memory-improvements.md:75-79` — the scoring formula. **Constraint:
+  untouched by this slice.**
+
+## §3 Objective
+
+Two active `decision` episodes in the same project that say nearly the same thing are stored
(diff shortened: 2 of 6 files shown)


lantiscooperdev (commented):
## Rule 17 bot review — PR #572 contradiction advisory (#537), round 1 ACCEPT

Covers the full arc: four parallel scout maps, a frozen on-disk build spec, a pi/MiniMax-M3 builder seat, an adversarial pi/GLM-5.2 frozen-diff reviewer, one fold round, and orchestrator-run verification of every claim below.

### Verdict trail

| Round | Scope | Reviewer | Blocking | Verdict | Reply artifact |
|---|---|---|---|---|---|
| 1 | Frozen diff, three layers (per-artifact → cross-file → whole-diff), 10 behavior-sim probes | pi/GLM-5.2 (neuralwatt), read-only seat | **0** | **ACCEPT** | `review-537-r1` |

Provider note: the second opinion ran on GLM-5.2 rather than the template default `codex`, on explicit operator instruction this session. The codex provider is additionally known-broken — it captures the seat's SessionStart bootstrap prompt as the reply body (#538), which is the next arc.

### Findings closed in commit 180dee5

| # | Finding | Class | Resolution |
|---|---|---|---|
| F4 | Cycle safety correct, but `testSupersedesChainLinkedPairExcluded` used a **linear** chain, so the `seen` guard was never the cause of termination — §13 EC5's coverage claim was overstated | test fidelity | Added `testChainCycleTerminates`: a genuine `a→b→a` cycle plus a self-referencing row. Without the guard this hangs rather than fails. |
| F6 | `MAX_GROUP` default not pinned — the existing test passed `{maxGroup: 5}`, so a regression to `Infinity` or `0` would have passed unchanged | falsifiability | Added `testDefaultMaxGroupIsPinned`: asserts the exported constant is 2000 and exercises the **default** cap with no override. |
| F8 | REQ-9's `data_dir` carriage confirmed present by simulation but asserted by no test | test fidelity | `testDoctorWarnsOnContradiction` now asserts `data_dir` is a non-empty string. |
| F9 | Spec §A.7 Step 1.1 verify said `5` exports; file has `6` | spec typo | Spec corrected; `grep -c "^export "` → `6` verified. |

### Re-verification (independently executed by the orchestrator, not taken from the seat)

- `node tests/test-contradiction.mjs` → **20 passed, 0 failed**. **PASS.**
- `node tests/test-contradiction.mjs --break-detector` → **exit 1, 16 passed / 4 failed**. Guard proven RED. **PASS.**
- `node tests/test-em-doctor.mjs` → 13/13. **PASS.**
- `node tests/test-relevance.mjs` → 17/17. **PASS.**
- `node tests/test-store-write-concurrency.mjs` → 46/46. **PASS.**
- `node tests/test-ci-suite-registration.mjs` → 16/16, 258 suites / 152 step-wired. **PASS.**
- `git diff --stat main -- scripts/lib/relevance.mjs` → empty. RFC-001 scoring contract byte-unchanged. **PASS.**
- F7 reachability challenged and disproven independently: `grep -c deprecated categories.json` → `0`. **PASS.**
- CI: all six checks green (`validate`, `p12-invariant-gate`, `substrate`, `gates-classifiers`, `recall-activation`, `clerk-consolidation`).

### Test posture (final)

20 tests in one new suite, wired into the `substrate` shard. 8 of the 20 are end-to-end runtime probes that spawn the **real** `em-store`, `em-revise`, and `em-doctor` against isolated fixture stores (`mkdtemp` cwd + `mkdtemp` HOME) and assert observed stdout / stderr / exit codes.

Both polarities are covered at every call site, which is what makes the detector meaningful rather than merely present:

- contradiction present → advisory fires, `em-doctor` reports `warn`
- unrelated summary, and wrong category → silence, `em-doctor` reports `ok`
- a real `em-revise` chain → `ok`; the sanctioned correction does **not** false-positive

The `--break-detector` argv flag raises the pure-layer threshold to an unreachable 1.1 so the positive assertions must go red. A guard never observed failing guards nothing.

### Safety posture

The high-blast-radius concern is that `em-store` is on every substrate write path. The advisory is placed after the write, after lock release, inside `try {} catch {}`, structurally mirroring the existing R9a collision report at `em-store.mjs:408-427`. REQ-7 asserts stdout stays exactly `{status,id,file,scope}` and exit stays 0 **while the advisory is firing** — verified at runtime, not by inspection. `loadIndex` itself swallows parse errors and returns `[]`, so a corrupt index degrades to "no advisory", never to a failed write.

Each carries all five DEFER fields — run-the-scenario, spec check, history check, same-class check, residual risk.

- **#568** — negation and short-subset lexical false positives. Same-class: filed with the other lexical-precision limits; every existing Jaccard consumer in the repo shares them.
- **#569** — non-ASCII summaries tokenize to empty, so identical CJK decisions are never flagged. Class-complete: the fix must land in `tokenizeQuery`, which `em-search`, `em-consolidate`, `em-promote`, and `em-topic-tracks` all share — fixing only this comparator would leave four siblings with the identical gap.
- **#570** — single-scope `chainLinked` resolution and strict category equality. The category half is verified unreachable today and filed as latent defense-in-depth.
- **#571** — `access_count` boost can entrench a stale unlinked decision over its revision terminal. This is #537's third part, which #537's own constraints section forbade fixing in an advisory slice; it needs an RFC-001 amendment and is filed with the sibling formula inputs (`feedback`, `pinned` floor) enumerated.

Per Rule 17, your approval in the UI is the `review-required` gate. The substantive review is complete — ACCEPT with 0 blocking findings, four test-fidelity findings fixed in a fold round, four dispositioned defers filed as issues, and CI green on all six checks. Worth a specific look from you: the 0.6 Jaccard threshold is a judgment call tuned against the #537 repro (which scores 0.667) rather than against a live corpus, so alert-fatigue is the risk most worth your eye. Please do a final read of the consolidated diff and approve, or HOLD with findings, in the GitHub UI.

🤖 Posted by lantiscooperdev (bot account) — Claude Opus 4.8 (1M context).

lantiscooperdev (approved):

PR #574, opened 2026-07-24
Gate provider replies before persistence (fixes #538)

lantisprime:
Closes #538.

## The bug

`scripts/second-opinion.mjs` persisted whatever a provider wrote to stdout as a review reply whenever the provider exited 0. The only pre-persistence checks were `r.timedOut` and `!r.ok` — exit-status shaped, never the body.

Issue #538's instance: a codex seat answered its own SessionStart bootstrap prompt, exited 0, and `Load session_handoff.md from 2026-07-14 16:32? (y/n)` was written as a `status: ok` review reply. A `--consensus` run would have counted it as a completed round. The single-dispatch path never ran `parseVerdict` at all, so it was entirely unguarded.

## The fix

New pure predicate `scripts/second-opinion/lib/reply-sanity.mjs`, called in `runDispatch` between the `!r.ok` check and `return r`. That is the single choke point both the consensus loop and the single-dispatch block pass through, so one edit closes both paths.

A reply is rejected when it is empty, whitespace-only, non-string, or below a 200-char floor **while carrying no `json:second-opinion-summary` fence**. Fence presence is checked *before* length, so a compliant terse review is never rejected for brevity. On rejection the raw stdout is preserved to `.review-store/forensics/<requestId>.round<N>.invalid-reply.txt` and the harness exits `provider-reply-invalid` with nothing persisted — mirroring the existing `provider-timeout` idiom. The floor is tunable with `--min-reply-chars`, validated in the pre-write block so an invalid value writes nothing.

Direction (b) from the issue was chosen over (a): it closes the class for every provider rather than one adapter's spelling, and it is testable through the stub provider.

## Verification

Every command below was run directly, not reported by a subagent.

| Claim | Observed |
|---|---|
| `node tests/test-second-opinion-reply-sanity.mjs` | `8 passed, 0 failed` |
| `node tests/test-second-opinion-reply-sanity.mjs --break-sanity` (negative control) | exit `1` |
| `node tests/test-so-reply-sanity-e2e.mjs` | `6 passed, 0 failed` |
| E2E suite with the gate deleted in a scratch copy | `2 passed, 4 failed`, exit 1 |
| `node tests/test-second-opinion-consensus-e2e.mjs` | `7 passed, 0 failed` |
| `node tests/test-second-opinion-dispatch.mjs` | `9 passed, 0 failed` |
| `node tests/test-second-opinion-providers.mjs` | `24 passed, 0 failed` |
| `node tests/test-ci-suite-registration.mjs` | `16 passed, 0 failed`, `PASS` |

Both new suites are step-wired into `tests.yml`; nothing was added to the shrink-only `KNOWN_UNWIRED` baseline.

Adversarial fail-closed probes against the built code, all exiting non-zero with zero reply records: empty stdout, whitespace stdout, non-string stdout (single and consensus), a forensics write that throws (`forensics` degrades to `null`, the gate still fires), and the verbatim #538 bootstrap on both dispatch paths.

## Review

Two rounds by a pi/GLM-5.2 seat, both **ACCEPT with 0 blockers**. Round 1 reviewed the plan and reproduced the original bug on a pristine copy before accepting; round 2 reviewed the frozen diff, confirmed it byte-identical to the frozen patch, and proved the E2E suite discriminates the gate by deleting it. Codex was not used as the provider (operator directive, 2026-07-23).

## Known residual and deferrals

- Long fence-less garbage at or above the floor still passes the gate. Out of scope: closing it means requiring the fence on every reply, a behavior change for every provider. Recorded in the plan's Open Decisions; no such instance is on record, so the 5-field defer discipline cannot be discharged today.
- `tests/test-second-opinion-storage.mjs` has one failing assertion — **pre-existing**, reproduced identically on unmodified `main`. Filed as #573, deliberately not fixed here to keep this diff scoped.

Plan and full disposition table: `docs/plans/issue-538-reply-sanity-gate.md`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -188,6 +188,12 @@ jobs:
       - name: Run second-opinion gate per-project install E2E (RFC-008 P4d — gate + libs + runbooks per-project; snapshot global)
         run: node tests/test-install-second-opinion-e2e.mjs
 
+      - name: Run reply-sanity predicate suite (#538 — pre-persistence gate on provider replies)
+        run: node tests/test-second-opinion-reply-sanity.mjs
+
+      - name: Run reply-sanity harness E2E (#538 — bootstrap-shaped reply rejected, nothing persisted)
+        run: node tests/test-so-reply-sanity-e2e.mjs
+
       - name: Run scripts-subtree deploy + completeness suite (#531 — fail-closed classification, UNDEPLOYED audit)
         run: node tests/test-install-scripts-subtrees.mjs
 
--- a/docs/plans/issue-538-reply-sanity-gate.md
+++ b/docs/plans/issue-538-reply-sanity-gate.md
@@ -268,15 +268,21 @@ Total: 14 tests. Runners:
 
 | Claim | Command (the strong-layer one) | Observed artifact |
 |---|---|---|
-| Predicate suite passes | `node tests/test-second-opinion-reply-sanity.mjs` | `<fill at build time>` |
-| Predicate guard is falsifiable | `node tests/test-second-opinion-reply-sanity.mjs --break-sanity` | `<must be non-zero exit>` |
-| E2E suite passes | `node tests/test-so-reply-sanity-e2e.mjs` | `<fill>` |
-| No regression in consensus E2E | `node tests/test-second-opinion-consensus-e2e.mjs` | `<fill>` |
-| No regression in dispatch suite | `node tests/test-second-opinion-dispatch.mjs` | `<fill>` |
-| No regression in storage suite | `node tests/test-second-opinion-storage.mjs` | `<fill>` |
-| CI registration lint green | `node tests/test-ci-suite-registration.mjs` | `<fill>` |
-| Deploys clean | `node tools/deploy-audit.mjs` | `<fill>` |
-| Merged | `gh pr view <n> --json state,mergeCommit` | `<fill>` |
+All rows below were run by the orchestrator directly, not reported by a seat.
+
+| Claim | Command (the strong-layer one) | Observed artifact |
+|---|---|---|
+| Predicate suite passes | `node tests/test-second-opinion-reply-sanity.mjs` | `8 passed, 0 failed` |
+| Predicate guard is falsifiable | `node tests/test-second-opinion-reply-sanity.mjs --break-sanity` | exit `1` (5 rejection assertions go red) |
+| E2E suite passes | `node tests/test-so-reply-sanity-e2e.mjs` | `6 passed, 0 failed` |
+| E2E suite discriminates the gate | reviewer deleted the gate in a scratch copy and reran | `2 passed, 4 failed`, exit 1 |
+| No regression in consensus E2E | `node tests/test-second-opinion-consensus-e2e.mjs` | `7 passed, 0 failed` |
+| No regression in dispatch suite | `node tests/test-second-opinion-dispatch.mjs` | `9 passed, 0 failed` |
+| No regression in providers suite | `node tests/test-second-opinion-providers.mjs` | `24 passed, 0 failed` |
+| Storage suite unchanged by this diff | `node tests/test-second-opinion-storage.mjs` on unmodified `main` AND on this branch | `10 passed, 1 failed` on BOTH — pre-existing, filed as issue #573 |
+| CI registration lint green | `node tests/test-ci-suite-registration.mjs` | `16 passed, 0 failed`, `test-ci-suite-registration: PASS` |
+| Deploys clean | `node tools/deploy-audit.mjs` | run post-merge; pre-merge drift (`MISSING=1 DIFFER=3`) is expected because the global install is older than the branch |
+| Merged | `gh pr view <n> --json state,mergeCommit` | `<fill at merge>` |
 
 ## §16 Risk Analysis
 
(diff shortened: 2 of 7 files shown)


lantiscooperdev (commented):
## Playbook v5 two-round adversarial review — PR #574 (#538 reply-sanity gate) ACCEPT

Covers the plan review and the frozen-diff review of the pre-persistence reply-sanity gate,
both behavior-simulated against isolated fixtures rather than read statically.

### Verdict trail

| Round | Scope | Reviewer | Blockers | Verdict | Artifact |
|---|---|---|---|---|---|
| 1 | Plan (`docs/plans/issue-538-reply-sanity-gate.md`) | pi / GLM-5.2 (neuralwatt) | 0 | **ACCEPT** | `scratchpad/review-plan-r1.md`, 441 lines |
| 2 | Frozen diff (`538-1`) | pi / GLM-5.2 (neuralwatt) | 0 | **ACCEPT** | `scratchpad/review-diff-r2.md`, 157 lines |

Codex was not used as the second-opinion provider per the operator directive of 2026-07-23.

### What round 1 established

The reviewer reproduced the original #538 failure on a pristine copy before accepting
anything: an unpatched harness plus a bootstrap-shaped stub stdout returned
`{"status":"ok", "reply":{"bodyPath":"…/replies/….body.md"}}` with the prompt text
persisted as a review. It then applied the plan's verbatim patches to a temp copy and drove
the real harness, confirming every §A.7 anchor matched exactly once against `main`.

### Findings closed before build (round 1)

| Finding | Sev | Class | Resolution |
|---|---|---|---|
| F-5 | P2 | Non-falsifiable verify | Step 2.2's verify was eyeball-only; a no-op passed it (proven: probe returned `status:ok`, exit 0). Replaced with a single `node -e` assertion that exits 1 on a no-op. The reviewer's suggested `\| grep -c` form was **not** taken — §A.6b bans pipes in a verify. |
| F-3 | P2 | False-positive surface | Terse legitimate *unfenced* replies under 200 chars are rejected. Deliberate per §5; documented as §13 EC8 naming `--min-reply-chars 0` as the operator escape hatch. |
| F-7 | P3 | Drift risk | `FENCE_RE` is duplicated from `consensus.mjs:36` with no sync test. Added `testFenceRegexInSyncWithConsensus` + §13 EC10. Import path corrected to `../scripts/…` (the suggested snippet used `../../`). |
| F-8 | P3 | Documentation | A rejected reply leaves the request record on disk. Documented as §13 EC9; identical to the pre-existing timeout path, so no code change. |
| F-2a | P3 | Defensive code | `reply-not-string` arm is unreachable for codex (always returns a string). REJECTED as a change — other providers are not bound by codex's implementation. |
| F-6 | P3 | Test-only surface | `SO_STUB_RAW_BODY` is test-only by convention, not enforcement. REJECTED as a change — pre-existing for every `SO_STUB_*` knob; no new production exposure. |
| F-4 | P3 | Residual | Long fence-less garbage still passes. DEFERRED per §5/§17 — see Known residual below. |

### Round 2 re-verification (independently executed)

- Staged `git diff --cached` byte-identical to the frozen patch. **PASS.**
- All three CREATE files equal §A.7.1/.2/.6 verbatim; all four EDIT blocks contiguous per §A.7.3/.4/.5/.7. **PASS.**
- Bootstrap reply, single-dispatch: exit 1, `provider-reply-invalid` / `reply-too-short-no-summary`, 0 replies, forensics verbatim. **PASS.**
- Bootstrap reply, consensus: exit 1, same code, `rounds=null`, 0 replies. **PASS.**
- Empty and whitespace-only stdout: exit 1, `reply-empty`, 0 replies. **PASS.**
- Non-string stdout (`undefined`), both paths: exit 1, `reply-not-string`, 0 replies, 0-byte forensics, no throw. **PASS.**
- Forensics write forced to throw (`forensics` pre-created as a file): exit 1, `forensics:null`, gate still fires, 0 replies. **PASS — a throwing forensics write does not open the gate.**
- Gate deleted in a scratch copy: E2E suite goes RED 2/6, failures quoting real captured output. **PASS — the suite discriminates the gate.**
- No third path to storage: `writeReplyRound` has exactly two call sites, both after `runDispatch` returns. **PASS.**

### Test posture (final)

- `test-second-opinion-reply-sanity.mjs` — `8 passed, 0 failed`; `--break-sanity` exits 1.
- `test-so-reply-sanity-e2e.mjs` — `6 passed, 0 failed`.
- `test-second-opinion-consensus-e2e.mjs` — `7 passed, 0 failed`.
- `test-second-opinion-dispatch.mjs` — `9 passed, 0 failed`.
- `test-second-opinion-providers.mjs` — `24 passed, 0 failed`.
- `test-ci-suite-registration.mjs` — `16 passed, 0 failed`, `PASS`.
- CI: all six checks green on this PR.

Both new suites are step-wired into `tests.yml`; nothing was added to the shrink-only
`KNOWN_UNWIRED` baseline.

### Known residual (accepted, not a blocker)

A fence-less reply at or above the 200-char floor still reaches storage: `status:ok` in
single-dispatch, written-then-`verdict-parse-failed` in consensus. Both rounds confirmed
this at runtime. Closing it means requiring the `json:second-opinion-summary` fence on every
reply, which is a behavior change for every provider and outside issue #538. Recorded in the
plan's §17 rather than filed, because the 5-field defer discipline's first field cannot be
discharged — no such instance is on record. The predicate is structured to grow a
`reply-fence-unparseable` arm if one appears.

- **#573** — `tests/test-second-opinion-storage.mjs:286` asserts a stale two-fragment
  default preamble list; the composer now returns three. Reproduced identically on
  unmodified `main` (`10 passed, 1 failed`), so it is **not** caused by this PR. Filed with
  all 5 defer fields. Root cause of its invisibility: all eleven `test-second-opinion-*.mjs`
  suites sit in the `KNOWN_UNWIRED` baseline, so no CI shard runs them.

Per Rule 17, your approval in the UI is the `review-required` gate. The substantive review
is complete — two independent rounds, both ACCEPT with zero blockers, every claim backed by
observed runtime output rather than code reading. Please do a final read of the consolidated
diff and approve (or HOLD with findings) in the GitHub UI.

🤖 Posted by lantiscooperdev (bot account) — Claude Opus 4.8 (1M context).

lantiscooperdev (approved):

PR #617, opened 2026-07-26
RFC-007 Phase 3: wiki-link + composes-with edges with slugify() resolution and dangling tracking (fixes #589)

lantisprime:
Implements RFC-007 Phase 3. Closes #589.

## What ships

Two rule-to-rule edge types on `em-graph.mjs`, both opt-in (registered in `EDGE_TYPES`, absent from `DEFAULT_EDGES`):

- **`wiki-link`** — `[[name]]` in rule bodies, fenced blocks and inline backticks excluded
- **`composes-with`** — top-level bullets under `## Composes with`, nested bullets ignored

Plus the shared target resolution both depend on, per-edge-type dangling tracking, and a 31-case suite wired into CI.

## The issue understated the work

#589 described two parsers and a fixture. Adding edges to rule nodes broke three assumptions Phase 2 had made explicit in its own comments:

| Site | Before | After |
|---|---|---|
| `--orphans` | reported **every** rule unconditionally | reports a rule only at degree 0 |
| `--from <rule-slug>` | early-returned a single node | falls through to the BFS, making good on the comment left at `em-graph.mjs:290` |
| `--hubs` | `degree` map was episode-only | rule ids seeded, so rules can rank |

Every line anchor #589 cited was also stale by +10 to +25 lines, drifted by Phase 2 (#604). Several pointed at blank lines; two pointed at the wrong JSON object entirely. All corrected.

## Target resolution needed a decision

Built literally, the spec resolves almost nothing. Measured against the live corpus before writing any code:

| Edge type | Spec as written | With filename-alias index |
|---|---|---|
| `wiki-link` | 13 of 132 | **122** of 132 |
| `composes-with` | **0** of 76 | **51** of 76 |

Node ids are `slugify(frontmatter name)` per `RFC-007:142`, but authors link by filename. `RFC-007:151` separately requires `composes-with` markdown links to resolve "by file basename", which needs an index `:142` disclaims. The RFC contradicted itself, functionally rather than cosmetically.

Resolution now tries the name-derived id first, then a filename-basename alias index; ambiguous basenames are dropped rather than guessed. **Node ids are unchanged and `:142` is untouched.** Documented in a new Target resolution subsection.

## Two RFC contradictions settled

- **Dangling bucket** is `wiki-link` everywhere. The prose said `wiki-link-dangling`, a spelling appearing nowhere else in the 737-line file and disagreeing with both the contract and the `kind` enum.
- **The wiki-link regex** was stated twice, non-equivalently. The prose form is canonical. Verified programmatically: the contract JSON-unescaped, the prose span, and the literal compiled into `em-graph.mjs` are now character-identical.

## Verification

| Claim | Command | Observed |
|---|---|---|
| New suite | `node tests/test-em-graph-links.mjs` | `31 passed, 0 failed` |
| Phase 1 regression | `node tests/test-em-graph.mjs` | `10 passed, 0 failed` |
| Phase 2 regression | `node tests/test-em-graph-nodes.mjs` | `28 passed, 0 failed` |
| Suite genuinely CI-wired | `node tests/test-ci-suite-registration.mjs` | `16 passed, 0 failed` |
| Negative control | `node tests/test-em-graph-links.mjs --break-wikilink` | exit 1, 11 failures |
| **REQ-8 default output** | worktree at `8e6e699`, `cmp` on `--hubs` and `--orphans` | **byte-identical** |
| RFC anchors | `sed -n '239p;251p;314p;320p;332p'` | all point where claimed |

REQ-8 was checked against the baseline commit directly rather than via the suite's own before/after comparison, which is the proxy layer one short of the claim.

**Live corpus:** 81 rule nodes carry `wiki-link` edges (9 dangling), 31 carry `composes-with` edges (1 dangling). `--from pi-minimax-m3-builder-seat --depth 2` returns 5 nodes and 4 edges where `8e6e699` returned `count: 1`.

## Review

Two rounds, both HOLD, all 12 findings dispositioned.

Round 1 on the plan caught a blocker: a step referenced a symbol a later step defined, so verbatim application would have thrown `ReferenceError`. Round 2 on the frozen diff caught a MAJOR: **CRLF line endings silently dropped every `composes-with` edge** — zero edges, zero dangling, exit 0. Latent on this corpus but the same silent-zero signature REQ-6 traps for, and inconsistent with `lib/rule-nodes.mjs:67` which already handles CRLF. Fixed with a regression test proven to fail without the fix.

Round 2 also caught a REQ-12 miss: the RFC's Scope "Unbuilt" list still enumerated Phase 2 and Phase 3.

## Pre-existing stalenesses fixed in passing

- The mermaid diagram labelled Phase 2 UNBUILT while the phases table and ledger both recorded it SHIPPED.
- `slugify._shipped_use` claimed the function was unused by the shipped script, though `scanRuleNodes` has called it since Phase 2.

`grep -c UNBUILT` on the RFC: 17 before, 10 after.

## Deliberately not included

**No Implementation ledger row.** Its provenance must come from `git log --follow` on the merge commit, which does not exist yet. Writing a branch-tip SHA there is the defect #582 was filed for and #605 had to repair. It will be filled post-merge.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -404,6 +404,9 @@ jobs:
       - name: Run em-graph node-projection suite (RFC-007 Phase 2)
         run: node tests/test-em-graph-nodes.mjs
 
+      - name: Run em-graph link-edge suite (RFC-007 Phase 3)
+        run: node tests/test-em-graph-links.mjs
+
       - name: Run em-doctor suite
         run: node tests/test-em-doctor.mjs
 
--- a/docs/plans/RFC-007-P3-wiki-link-composes-with.md
+++ b/docs/plans/RFC-007-P3-wiki-link-composes-with.md
@@ -257,7 +257,7 @@ empty strings. Never null.
 | EC2 | `[[Name]]` vs `[[name]]` in two files | both slugify to `name`, one deduped edge | `t_case_folded_targets_dedupe` |
 | EC3 | Slug collision already exits 2 at `:152-155` before any body scan | Phase 3 never runs on a colliding corpus | `t_collision_still_exits_2_before_extraction` |
 | EC4 | `## Composes with` present, section empty | 0 edges AND 0 dangling (REQ-6) | `t_composes_empty_section_no_dangling` |
-| EC5 | Unclosed fence — ` ``` ` opened and never closed | remainder of file treated as fenced, no edges (matches `cites` regex behavior) | `t_unclosed_fence_swallows_remainder` |
+| EC5 | Unclosed fence — ` ``` ` opened and never closed | **remainder is NOT treated as fenced**; a wiki-link after it IS extracted. `stripFenced` is non-greedy and requires a closing fence, which is exactly what shipped `cites` does at `em-graph.mjs:207`. Parity with `cites` is the requirement (REQ-4), so this is the correct behavior, not a gap | `t_unclosed_fence_does_not_swallow_remainder` |
 | EC6 | `## Composes with` followed immediately by `## Other` | section is empty, next heading terminates it | `t_composes_terminated_by_next_heading` |
 | EC7 | Live corpus run — 68 files carry `[[…]]`, 21 carry the heading | real run completes, emits both edges and dangling entries | `manual:` §15 row |
 
@@ -274,7 +274,7 @@ Group 1: wiki-link shape matrix (15)
 Group 2: exclusion (3)
   t_wikilink_fence_exclusion
   t_wikilink_inline_backtick_exclusion
-  t_unclosed_fence_swallows_remainder
+  t_unclosed_fence_does_not_swallow_remainder   (EC5, corrected: parity with cites)
 
 Group 3: composes-with (8)
   t_composes_toplevel_only
--- a/docs/plans/RFC-007-P3-wiki-link-composes-with.md
+++ b/docs/plans/RFC-007-P3-wiki-link-composes-with.md
@@ -1,0 +1,776 @@
+# RFC-007-P3 — `wiki-link` + `composes-with` edges Plan
+
+## §1 Status
+
+`Planning only.` Do not implement until this plan and its review are accepted (Rule 18).
+Current stage: **round-1 review folded, awaiting final approval**. Round 1 returned HOLD with
+7 findings; all 7 are dispositioned in §19.1 and the 6 actionable ones are applied.
+
+| Field | Value |
+|---|---|
+| RFC | `RFC-007` (graph projection) |
+| Parent requirements | Phase 3 row `RFC-007-graph-projection.md:384`; test plan `:421-425` |
+| Workplan episode | `20260725-143851-workplan-v268-587-merged-at-8e6e699-with-687e` (v268) |
+| Target branch | `feat/rfc-007-p3-wiki-link-composes-with` |
+| Executor altitude (§0.1) | **low** — Appendix A is the build path |
+| GitHub issue | #589 |
+
+## §2 Episode Search Summary
+
+```bash
+node scripts/em-search.mjs --tag rfc-007 --scope all --limit 12 --no-track --no-score
+```
+
+Returned 3 rows, none carrying design constraints beyond the workplan/handoff pair:
+
+- `20260725-143851-…-687e` (workplan v268): queue head is #589; `main` = `8e6e699`; no open PRs.
+- `20260725-143935-…-7a63` (handoff 93): Phase 2 and #587 both touched `em-graph.mjs`; the
+  surface is warm. Carries "never accept a seat's verify summary" — the builder seat on the
+  #587 arc reported 9 verifies matched spec and silently omitted the one that failed.
+- `20260722-133545-…-99bf` (playbook v15, pinned, `authoritative`): pipeline order and role lock.
+
+Verified-still-present (memories are point-in-time): `slugify()` at `scripts/lib/rule-nodes.mjs:30-38`
+and `scanRuleNodes` imported at `scripts/em-graph.mjs:48` — both confirmed by direct read, not recall.
+
+## §3 Objective
+
+`em-graph.mjs` projects two new rule-to-rule edge types from rule-file bodies — `wiki-link`
+(`[[name]]`) and `composes-with` (top-level bullets under `## Composes with`) — resolved to rule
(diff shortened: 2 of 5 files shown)


lantiscooperdev (commented):
## Tiered multi-agent review — RFC-007 Phase 3, two rounds, ACCEPT

Covers the full #617 diff (5 files, +1638/-34) across a scout stage, a plan review and a frozen-diff review. Both review rounds returned HOLD; all 12 findings are dispositioned and the actionable ones applied.

### Verdict trail

| Round | Scope | Seat | Verdict | Blockers |
|---|---|---|---|---|
| Scouts | RFC anchors, `em-graph.mjs` runtime state, test/CI wiring, governance + disposition sweep | 4 parallel Sonnet | n/a | n/a |
| 1 | The plan, with Appendix A frozen | pi / neuralwatt glm-5.2 | HOLD → folded | 1 BLOCKER, 2 MAJOR, 4 MINOR |
| 2 | Frozen diff `main...HEAD` | pi / neuralwatt glm-5.2 | HOLD → fixed in `f76dc6b` | 1 MAJOR, 4 MINOR |

Seat spend for the whole arc: **$2.71**. No Fable seats, no openrouter.

### Round 1 findings, closed before any code was written

| # | Class | Resolution |
|---|---|---|
| 1 | BLOCKER | Step 1.8's payload called `danglingOut()`, defined by step 1.9. Verbatim application would have thrown `ReferenceError` and dropped Phase 2 to 27/28 with no clean in-step fix. Stray call removed. |
| 2 | MAJOR | Step 1.4's verify asserted a `dangling` key nothing emits until 1.9. Rewritten to a loop-presence grep. |
| 3 | MAJOR | Step 1.9 expected `grep -c 'danglingOut()'` → 4 counting the definition, but the definition reads `danglingOut = ()` so the substring never matches it. Corrected to 3. |
| 4 | MINOR | Step 3.5's relative UNBUILT count was ambiguous. Restated absolutely: 17 at `8e6e699`, 15 after 3.2, 10 after 3.5. |
| 5 | MINOR | `composesTargets` slugified whole prose bullets into garbage refs. Now yields no target. |
| 6 | MINOR | Two backticked spans on one bullet: first wins, made explicit. |
| 7 | MINOR | Presence-only greps, assessed as back-stopped by the functional suites. No change. |

### Round 2 findings, closed in `f76dc6b`

| # | Class | Resolution |
|---|---|---|
| 1 | **MAJOR** | **CRLF line endings silently dropped every `composes-with` edge** — zero edges, zero dangling, exit 0. The heading matched (that test trims) but the bullet kept a trailing `\r` and the bullet regex did not match. Same silent-zero signature REQ-6 traps for, and inconsistent with `lib/rule-nodes.mjs:67` which already handles CRLF. Fixed by splitting on `/\r?\n/`. |
| 2 | MINOR | A markdown-link target with a `#fragment` slugified the fragment into the target. Stripped before basename. |
| 3 | MINOR | Section start trimmed, terminator did not, so an indented `## Composes with` could start a section an indented `## Next` could not end. Both now test the trimmed line. |
| 4 | MINOR | RFC Scope's "Unbuilt" list still enumerated Phase 2 and Phase 3, contradicting three other places in the same file. A REQ-12 miss: the first sweep only covered the sites the plan enumerated. |
| 5 | MINOR | The two new contract entries lacked the `shipped_at` field all five older entries carry. |

### RFC contract changes

- Edge-status rows `:84-85` and both contract `status` fields → SHIPPED, with verified anchors.
- Dangling bucket unified to `wiki-link`; the `wiki-link-dangling` spelling appeared nowhere else in the file.
- Machine-readable regex rewritten to match the prose form. **Verified programmatically**: the contract JSON-unescaped, the prose span, and the literal compiled into `em-graph.mjs` are character-identical.
- New **Target resolution** subsection defining the filename-alias index, reconciling the "resolved by file basename" clause with the frontmatter-name node-id rule. `:142` itself untouched.

### Orchestrator re-verification (independently executed, not taken from seat reports)

- Suites: links **31/31**, phase1 **10/10**, phase2 **28/28**, ci-suite-registration **16/16**. **PASS.**
- Negative control `--break-wikilink` → exit 1, 11 failures, and the passing tests are exactly the ones not dependent on wiki-link extraction. **PASS.**
- **REQ-8 against the real baseline**: worktree at `8e6e699`, `cmp` on `--hubs` and `--orphans` output → byte-identical. The suite's own before/after check is the proxy layer one short of this. **PASS.**
- RFC anchors `239, 251, 314, 320, 332` each verified to point where claimed, re-checked after the S4 line shifts. **PASS.**
- CRLF guard proven non-vacuous: the bullet regex returns `false` against a trailing `\r`, so the new test fails without the fix. **PASS.**
- Payload fidelity: every §A.5 block diffed against the plan, verbatim including comments. **PASS.**

### Live corpus behavior

| Edge | Nodes with edges | Dangling |
|---|---|---|
| `wiki-link` | 81 of 127 rule nodes | 9 |
| `composes-with` | 31 | 1 |

`--from pi-minimax-m3-builder-seat --depth 2` returns 5 nodes and 4 edges; `8e6e699` returned `count: 1`.

- **#618** — fill the Implementation ledger row post-merge from `git log --follow`. Deliberately omitted here: the provenance SHA does not exist until merge, and a branch-tip SHA is the defect #582 was filed for and #605 had to repair. Carries all 5 DEFER fields.
- **#619** — `em-console.mjs:114` flag whitelist omits `--nodes`/`--edges`, so rule/rfc nodes and Phase 3 edges are unreachable through the console. Pre-existing since Phase 2, recorded as an explicit non-goal.
- **#620** — an unclosed code fence does not suppress extraction, affecting shipped `cites` as well as new `wiki-link`. Design question surfaced by the build, deliberately not decided inside this slice.

### CI

8/8 green: `validate` ×3, `substrate`, `gates-classifiers`, `p12-invariant-gate`, `recall-activation`, `clerk-consolidation`.

Per Rule 17, your approval in the UI is the `review-required` gate. Substantive review is complete and both rounds are closed. Worth a specific look before approving: the **filename-alias index** is a genuine spec extension, adopted because the literal spec resolved 13/132 wiki-links and 0/76 composes-with bullets against the live corpus. It does not change node ids. If you disagree with that call, it is the one decision here that is architectural rather than mechanical.

🤖 Posted by lantiscooperdev (bot account) — Claude Opus 5 (1M context).

lantiscooperdev (approved):

